PR #5, opened 2016-03-11
Add study view pie chart component

zhx828:
- Migrate partial study view pie chart code (color schema included)
- Fix the dragging unstable issue


zhx828:
@yichaoS could you take a look the code and merge it? Thank you.

PR #16, opened 2016-03-28
Enable sharing Virtual cohort

kalletlak:
Share virtual cohort as a link.
Generated link example : localhost:9002/?vc_id=56f98d73d4c601ac2ef114f1.

### DIFF
--- a/app/scripts/views/components/editableRowComponent.js
+++ b/app/scripts/views/components/editableRowComponent.js
@@ -76,19 +76,6 @@
         shortenedLink: '---'
       };
     },
-    ready: function() {
-      var classname = document.getElementsByClassName('btn-share');
-      if (clipboard !== null) {
-        clipboard.destroy();
-      }
-      clipboard = new Clipboard(classname);
-      clipboard.on('success', function(e) {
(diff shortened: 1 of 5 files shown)


app/scripts/views/components/editableRowComponent.js
```diff
@@ -47,20 +51,44 @@
...
       return {
-        edit: false
+        edit: false,
+        share: false,
+        shortenedLink: '---'
       };
     },
+    ready: function() {
+      var classname = document.getElementsByClassName('btn-share');
+      if (clipboard !== null) {
```
zhx828: maybe use 'clipboard instanceof Clipboard'

app/scripts/views/components/editableRowComponent.js
```diff
@@ -47,20 +51,44 @@
...
     ],
     data: function() {
       return {
-        edit: false
+        edit: false,
+        share: false,
+        shortenedLink: '---'
       };
     },
+    ready: function() {
```
zhx828: Need to debug a little bit. This function will be called at least three times in one click.

Issue #91, opened 2016-07-08
Add case selection button


yichaoS:
-- radio buttons: patient id vs. sample id

PR #95, opened 2016-07-10
Add Case selection box

kalletlak:
Fix issue #91


yichaoS:
screen shot 2016-07-11 at 10 56 22 am
screen shot 2016-07-11 at 11 00 58 am

yichaoS:
I think it works fine. 
Just some small styling issues. [See screen shots]
The box in the popup is somehow wider than the window itself.
And "sample ID" and "patient ID" radio box is all jammed. Maybe some space between looks better.
And there's a apostrophe behind the "Custom" filter.

PR #135, opened 2016-07-25
Update download functions

zhx828:
1. Add download for tables
2. Change the download structure to only initialize/update download data after clicking the related button


app/scripts/views/components/dataTable/tableView.js
```diff
@@ -78,7 +80,9 @@
 
       return result;
     }
-    
+
+    content.dataForDownload = {};
```
kalletlak: I think dataForDownload is initialized in generalChart.js, then why to initialize in every chart implementation?

PR #141, opened 2016-07-28
Fix issue #103

kalletlak:


app/scripts/vueCore.js
```diff
@@ -131,6 +133,114 @@
              };
              this.charts = {};
              },*/
+            openCases:function(){
+              iViz.openCases();
+              
+             /* var studyId = '';
```
zhx828: Can we remove these commented out code? I think they have been isolated into an individual function?

app/scripts/vueCore.js
```diff
@@ -262,6 +372,21 @@
...
             });
+            $('.study-view-header-left-5').qtip({
+              content: {text: 'Click to view the selected cases' },
+              style: { classes: 'qtip-light qtip-rounded qtip-shadow' },
+              show: {event: 'mouseover'},
+              hide: {fixed:true, delay: 100, event: 'mouseout'},
+              position: {my:'bottom center', at:'top center', viewport: $(window)}
+            });
+            $('#study-view-header-left-6').qtip({
+              prerender: true,
```
zhx828: Why prerender the qtip?

app/styles/partials/_global.scss
```diff
@@ -72,14 +72,14 @@
 }
 
 #study-view-header-right-1 {
```
zhx828: Should we get rid of 'study-view' name?

kalletlak:
Created another PR #147 to accommodate same changes.

PR #142, opened 2016-07-28
Separate session module from iViz

kalletlak:


kalletlak:
Create another PR #143, with similar changes.

PR #145, opened 2016-08-02
Remove session-module related files

kalletlak:
Included them as single library 
1. vc-session.js
2. vc-session-vendor.js
3. vc-session.css
4. vc-session-vendor.css


app/scripts/vueCore.js
```diff
@@ -339,5 +316,5 @@
         return vmInstance_;
       }
     };
-  })();
+  })();*/
```
yichaoS, line 332: Maybe we just delete these lines

bower.json
```diff
@@ -5,7 +5,7 @@
     "bootstrap": "^3.3.6",
     "chosen": "~1.4.2",
     "datatables": "#1.10.2",
-    "dcjs": "~2.0.0",
+    "dcjs": "2.0.0-beta.31",
```
yichaoS, line 8: Why we are upgrading it to beta?

app/scripts/vueCore.js
kalletlak, line 332: We can, but just want to keep it for future reference.

bower.json
kalletlak, line 8: When you delete the bower_components folder and run the 'bower install' command, it would download 2.0.0-beta.32 version. With this version we would be running into issue in function [filterLabels()](https://github.com/cBioPortal/iViz/blob/multi-cohort/app/scripts/views/components/pieChart/pieChart.js#L266) because the svg structure generated is different compared to the one generated to dcjs 2.0.0-beta.31 version.

PR #146, opened 2016-08-03
Performance

kalletlak:
1. Improve performance.
2. Code cleanup
   - Add dimension for each special chart.
   - Use dimension filtered data instead of looping to all data and applying filters to sync between charts.
   - Show loading image for special charts when filters are updated. 
   - Use dimension data to update special charts.


app/scripts/controller/sync.js
```diff
@@ -71,6 +70,16 @@
     });
     return _selectedCasesData;
   };
+
+  iViz.sync.selectByCases = function(type_, data_, cases_) {
+    var caseIndices = iViz.getCaseIndices(type_);
+    var resultData_ = [];
+    $.each(cases_, function(key,val){
```
zhx828: Can we use _.each for the performance and consistency?

kalletlak: Sure! Done.

PR #147, opened 2016-08-04
Add query buttons

kalletlak:


app/index.html
```diff
@@ -110,29 +110,44 @@
...
           <span id="queryByGeneTextSpan"></span>
-          <textarea id="query-by-gene-textarea" class="expand expandFocusOut" rows="1" cols="10"></textarea>
-        </span>
-      </div>
+          <textarea id="query-by-gene-textarea" class="expand expandFocusOut" 
+                    :class="{'iviz-button-hover':highlightAllButtons}" rows="1" cols="10"></textarea>
+      </span>
+        <img id="arrow_studyview" src="images/arrow_studyview.png" alt="">
+        <form id="iviz-form"  v-on:submit.prevent="submitForm" method="post" 
+              action="http://www.cbioportal.org/index.do" target="_blank" style="float: left;">
```
yichaoS: Can we leave this as a parameter instead of hard-coded here.

kalletlak: Good catch! We can do that.

yichaoS: Let's not hard coded this.

PR #157, opened 2016-08-10
Change initLabel function in pieChart; Remove unnecessary initialization

zhx828:
No more DOM manipulation


kalletlak:
Mouseover on a row in the table is not highlighting the correct pie in pie chart.

PR #160, opened 2016-08-11
Dynamically updating group crossfilter data

kalletlak:
1. Add Attribute data to group cross filter dynamically
2. Code cleanup


kalletlak:
Implemented using PR #193

Issue #152, opened 2016-08-05
Include customization parameters in index.html as part of initialization parameter


zhx828:

zhx828:
Hide session buttons etc.

Issue #177, opened 2016-08-24
Migrate the latest features from study view


zhx828:

zhx828:
Reopen because pull request #212

Issue #79, opened 2016-07-01
set priority for showing charts


yichaoS:

zhx828:
Survival > Scatter > Tables > Pie(Cancer Type, Cancer Type Details)

zhx828:
We may need to assign different priority even within same chart type (edited)

Take MSKIMPACT as example, we want to show Cancer Type and Cancer Type Details as first two charts, but they are pie charts.

If we only assign same priority within same type, they will never be treated as first two

PR #178, opened 2016-08-25
Issues fixed for first milestone

### LINKED ISSUE #149: Abstract dimension implementation for regular/log scaled bar chart

zhx828:
There are known issues in cBioPortal: https://github.com/cBioPortal/cbioportal/issues/1622, will include these fixes after the issue solved.

### DIFF
--- a/app/scripts/controller/util.js
+++ b/app/scripts/controller/util.js
@@ -33,6 +33,8 @@
 'use strict';
 (function(iViz, _, cbio) {
   iViz.util = (function() {
+    var clinicalAttrsPriority = ['cancer_type', 'cancer_type_detailed',
+      'gender', 'age', 'sequenced', 'has_cna_data', 'sample_count_patient']; // DESC, all lowercase
     var content = {};
 
     /**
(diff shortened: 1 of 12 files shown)


zhx828:
Fix #149: abstract regular/log dimension; 
Add outlier features: #177;
Fix #152
Fix #139

zhx828:
The above commit https://github.com/cBioPortal/iViz/pull/178/commits/fe301e449a71bec63153f08c6fe05f5e8be884a9 is related to #79

app/scripts/controller/util.js
```diff
@@ -630,6 +632,158 @@
...
+     * Compare based on data availability.
+     * Notice that: attribute with only one category will be moved to end.
+     * Number of keys in this attribute is more than half numOfDatum
+     * will be moved to end as well.
+     *
+     * @param {object} a Attribute meta item A.
+     * @param {object} b Attribute meta item B.
+     * @return {number} Indicator which item is selected.
+     */
+    content.compareClinicalAvailability = function(a, b) {
```
kalletlak, line 686: In the next milestone as we are thinking of lazy loading to clinical data right, then how would we use this function? I mean we won't be having 'numOfDatum' info initially while loading right.

app/scripts/model/dataProxy.js
```diff
@@ -340,28 +357,39 @@
...
             _osSurvivalAttrMeta.show = true;
-            _patientAttributes.unshift(_osSurvivalAttrMeta);
+            _osSurvivalAttrMeta.priority = 1;
+            _patientAttributes[_osSurvivalAttrMeta.attr_id] = _osSurvivalAttrMeta;
           }
 
           // add Cancer Study
           if (self.getCancerStudyIds().length > 1) {
-            _patientAttributes.unshift({
+            _patientAttributes['study_id'] = {
```
kalletlak, line 421: Should we add keys, numOfDatum, priority attributes to this meta object and even missing keys, numOfDatum to the above charts(tables survival and scatter meta).

app/scripts/model/dataProxy.js
```diff
@@ -340,28 +357,39 @@
...
+            }
+          });
+          
           _result.groups = {};
           _result.groups.patient = {};
           _result.groups.sample = {};
           _result.groups.group_mapping = {};
-          _result.groups.patient.attr_meta = _patientAttributes;
-          _result.groups.sample.attr_meta = _sampleAttributes;
+          _result.groups.patient.attr_meta = iViz.util.sortClinicalAttrs(_.values(_patientAttributes));
```
kalletlak: Instead of applying 'sortClinicalAttrs' again why don't we use 'priority' attribute in each meta attribute to sort them?

app/scripts/controller/util.js
zhx828: Probably we need to calculate the priority every time we load additional attribute.

app/scripts/model/dataProxy.js
zhx828: I thought about it. Not really necessary. It's nice to have but it needs couple few loops. These attributes are mostly designed specifically, and normally they are in preSelectedAttribtues or in clinicalAttrsPriority list. And I have https://github.com/cBioPortal/iViz/pull/178/commits/fe301e449a71bec63153f08c6fe05f5e8be884a9#diff-5baab798025ed50b358d0b2e1421cbe3R687 to deal with attributes without keys or numOfDatum

app/scripts/model/dataProxy.js
zhx828: Good idea, will modify that.

PR #184, opened 2016-08-29
Code clean up based on Google style and eslint rules

zhx828:


app/scripts/controller/sync.js
```diff
@@ -1,67 +1,33 @@
...
 /**
  * @author suny1@mskcc.org on 3/15/16.
  */
 
 'use strict';
-(function(iViz, $, _){
+(function(iViz, $, _) {
   iViz.sync = {};
   // syncing util: select samples or patients based on only samples/patients filters
   iViz.sync.selectByFilters = function(filters, data) { // type: sample or patient
```
kalletlak: I think this function is not used anymore. We can get rid of it.

app/scripts/controller/sync.js
```diff
@@ -74,12 +40,12 @@
   iViz.sync.selectByCases = function(type_, data_, cases_) {
```
kalletlak: Even this function is not used anymore. We can get rid of this complete file.

zhx828:
Issues not addressed in the code reformat:
1. main.js _data doesn't have any reference
2. line 276 has http://eslint.org/docs/rules/no-loop-func issue
3. Issues with proxy.js under survivalChart folder

zhx828:
@yichaoS Could you take a look all survival plots related files? Are we going to use these files in iViz or in cBioPortal?

PR #193, opened 2016-08-31
Lazy loading clinical data

kalletlak:
1. Fetch clinical data using api when the chart is shown in dashboard
2. Update 'clear filters' logic


zhx828:
Pie table selected rows are not unchecked after clicking 'Clear All' button. portal-rc doesn't have this issue.

PR #195, opened 2016-09-02
Fix #191

### LINKED ISSUE #191: Lazy loading MutatedGene/CNA data

kalletlak:


zhx828:
Found an issue while testing on this. May not related to this pull. Please see screenshot.
The sample with 0 CNA is not selected. @yichaoS could you take a look?

screen shot 2016-09-03 at 10 20 44 pm

app/scripts/main.js
```diff
@@ -99,10 +95,6 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
       group.hasfilters = false;
       _.each(data_.groups.sample.attr_meta, function(attrData) {
         attrData.group_type = group.type;
-        if (attrData.view_type === 'table') {
```
zhx828, line 102: Just curious, why we have this special condition before?

kalletlak: This to separate table data from the vue modal. This is done to reduce the number of reactive data components.

kalletlak:
I think by default we are showing all samples in scatter plot even though if it doesn't have CNA data. May be we should hide those samples in scatter plot.

PR #200, opened 2016-09-06
Fix issue #198 & #199

### LINKED ISSUE #199: Scatter plot issue : cases with no cna data
Currently it is set to 0 when there is no cna data.

b221d906-7224-11e6-94ac-cf902a744162

kalletlak:


app/scripts/views/components/scatterPlot/scatterPlot.js
```diff
@@ -9,6 +9,17 @@
     var chartId_;
     var data_;
     var opts_;
+    var getQtipString = function(_data) {
+      var toReturn = 'Cancer Study:' + _data.study_id + '<br>Sample Id: ' +
+        _data.sample_id + '<br>CNA fraction: ';
+      if (isNaN(_data.cna_fraction)) {
```
zhx828, line 15: The situation is if the data isNaN, the case will not even be shown in scatter plot. It's ok to have this condition, but I don't think it will be used.

kalletlak: This is what the current cbioportal has. Cases with 'NA' cna or mutation count data are not shown.

Issue #197, opened 2016-09-02
Make table height ajustable based on number of categories


zhx828:
Attribute with less than 4 categories should use 'grid-h-1'

Issue #188, opened 2016-08-30
Show attributes with more than half number of patients/samples categories into tables


zhx828:
For example in hnsc_tcga
![screen shot 2016-08-30 at 4 13 02 pm](https://cloud.githubusercontent.com/assets/5400599/18105236/a9d47264-6ecc-11e6-89f1-694a38e60da0.png)
![screen shot 2016-08-30 at 4 13 08 pm](https://cloud.githubusercontent.com/assets/5400599/18105244/b063d23c-6ecc-11e6-9ee7-0ede26cf811f.png)

This change should only apply to attribute with STRING as data_type

PR #211, opened 2016-09-08
Visualize STRING data in table view if number of categories more than half size of study

zhx828:
This solves issues #176 #188 #194 #197


app/scripts/main.js
```diff
@@ -236,28 +236,69 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+        window.iviz.datamanager.getClinicalData(attrIds, isPatientAttributes))
+        .then(function(clinicalData) {
+          var _caseIdToClinDataMap = {};
+          var idType = isPatientAttributes ? 'patient_id' : 'sample_id';
+          _.each(clinicalData, function(_clinicalAttributeData, _attrId) {
+            var attrs = isPatientAttributes ? data_.groups.patient.attr_meta :
+              data_.groups.sample.attr_meta;
+            var selectedAttrMetaIndex = -1;
+
+            _.every(attrs, function(attr, index) {
```
kalletlak: You can use charts variable to get attribute object(something like  charts[_attrId]) instead of looping all attributes.

kalletlak:
Should we hide table display name for these pietables. They seems to be redundant with first column name.

app/scripts/views/components/dataTable/tableView.js
```diff
@@ -143,21 +156,110 @@
...
+        _.each(_selectedMap, function(datum) {
+          if (_.isArray(datum)) {
+            numOfCases += datum.length;
+          }
+        });
+      } else {
+        numOfCases = data_.length;
+      }
+
+      if (hasSelectedCases) {
```
kalletlak: Any specific reason to have similar conditional statement and looping _selectedMap (lines 205,206 and 215,216)?

app/scripts/views/components/dataTable/tableViewTemplate.js
```diff
@@ -48,20 +51,27 @@
...
       },
       'update-special-charts': function() {
-        var attrId =
-          this.attributes.group_type === 'patient' ? 'patient_id' : 'sample_id';
-        var _selectedCases =
-          _.pluck(this.invisibleDimension.top(Infinity), attrId);
-        this.chartInst.update(_selectedCases, this.selectedRows);
-        this.setDisplayTitle(this.chartInst.getCases().length);
-        this.showLoad = false;
+        // Do not update chart if the selection is made on itself
```
kalletlak, line 63: I remember that we though of updating special tables(Mutated genes and CNA tables) on submit operation (this is basically for updating the Freq column). May be we need to discuss about this.

PR #223, opened 2016-09-12
Fix eslint linting errors 

zhx828:


app/scripts/main.js
```diff
@@ -42,6 +42,23 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+    var str = [];
+    _.each(attrs, function(displayName, attrId) {
+      if (attrId === 'cna_details' || attrId === 'mutated_genes') {
+        var temp = 'No';
+        if (arr[attrId] !== undefined) {
+          temp = arr[attrId].length > 0 ? 'Yes' : 'No';
+        }
+        str.push(temp);
+      } else {
+        str.push(arr[attrId]); // This may return undefined. Is this expected?
```
kalletlak: Should be check for undefined and if yes shall we set it to 'NA'?

PR #230, opened 2016-09-12
Add study id attribute for all cases data

kalletlak:
This would be useful preparing hyperlink for samples/patients


kalletlak:
This would be useful to fix #229

Issue #260, opened 2016-09-15
$.get vars.json may return string


zhx828:
Looks like this only happens in web-app container. Or maybe related to tomcat configuration.

Issue #256, opened 2016-09-15
Patient Gender is always converted to Pie Table


zhx828:
Study: lgg_ucsf_2014

Issue #255, opened 2016-09-15
Cancer Type/Cancer Type Details should only be prioritized in MSK-IMPACT study


zhx828:
Test study: lgg_ucsf_2014

Issue #254, opened 2016-09-15
Lowercase na


zhx828:
acc_cmo_reidyd_06313_b
![screen shot 2016-09-15 at 10 13 20 am](https://cloud.githubusercontent.com/assets/5400599/18553056/0ed6ad32-7b2d-11e6-82d7-f6abeb8fff63.png)

PR #261, opened 2016-09-15
Fix issues

### LINKED ISSUE #246: Remove all CopyRights in js files

zhx828:
Fixed #246 #253 #254 #255 #256 #260

### DIFF
--- a/app/scripts/model/dataProxy.js
+++ b/app/scripts/model/dataProxy.js
@@ -17,6 +17,16 @@ window.DataManagerForIviz = (function($, _) {
   };
 
   /**
+   * Priorities (number before 10 are reserved for manually assign)
+   * TODO: priority manager after 10
+   *
+   * Survival Plots: 1; Scatter Plot: 2;
+   * Mutated Genes table: 3; CNA table: 4;
+   * Mutated Count bar chart: 5; CNA bar chart:6
+   * Cancer Type/Cancer Type details: 4.1 but MSKIMPACT: 0.9
+   */
+
+  /**
    * General pick clinical attributes based on predesigned Regex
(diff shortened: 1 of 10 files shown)


app/scripts/model/dataProxy.js
```diff
@@ -298,8 +308,14 @@ window.DataManagerForIviz = (function($, _) {
                     sequencedCases: _caseLists.allSampleIds
                   };
                 }
-                if (['CANCER_TYPE', 'CANCER_TYPE_DETAILED'].indexOf(_metaObj.attr_id) !== -1) {
-                  _metaObj.priority = 0.9;
+                if (['CANCER_TYPE', 'CANCER_TYPE_DETAILED']
+                    .indexOf(_metaObj.attr_id) !== -1) {
+                  if (Object.keys(_studyToSampleToPatientMap)
+                      .indexOf('MSKIMPACT') === -1) {
+                    _metaObj.priority = 4.1;
```
kalletlak, line 315: These priorities are hard coded( Changing priority depending on the study). May be we need to come up with a better solution in future release.

app/scripts/views/components/dataTable/tableView.js
```diff
@@ -227,11 +227,15 @@
       }
       return assignedColors[key];
     }
+
     function initPieTableData() {
       _.each(group.all(), function(attr, index) {
+        if (!attr.value || attr.value === 'na') {
```
kalletlak: Instead of checking it here why don't we do it at the time of creating dimension.

app/scripts/model/dataProxy.js
zhx828: Agree.

PR #281, opened 2016-09-19
Update dataproxy

kalletlak:
Set patient list when sample list is passed in the input StudyCasesMap

### DIFF
--- a/app/scripts/model/dataProxy.js
+++ b/app/scripts/model/dataProxy.js
@@ -927,7 +927,7 @@ window.DataManagerForIviz = (function($, _) {
                 }
                 // set patient list in studyCasesMap if sample list is
                 // passed in the input
-                if (_studyCasesMap[cancer_study_id].samples !== undefined && _studyCasesMap[cancer_study_id].samples.length > 0) {
+                if (_.isArray(_studyCasesMap[cancer_study_id].samples)) {
                   self.studyCasesMap[cancer_study_id].patients = _.unique(patientList);
                 }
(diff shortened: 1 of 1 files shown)


app/scripts/model/dataProxy.js
```diff
@@ -914,8 +915,15 @@ window.DataManagerForIviz = (function($, _) {
...
               }).then(function(data) {
                 var sample_to_patient = {};
+                var patientList = [];
                 for (var i = 0; i < data.length; i++) {
                   sample_to_patient[data[i].id] = data[i].patient_id;
+                  patientList.push(data[i].patient_id);
+                }
+                // set patient list in studyCasesMap if sample list is
+                // passed in the input
+                if (_studyCasesMap[cancer_study_id].samples !== undefined && _studyCasesMap[cancer_study_id].samples.length > 0) {
```
zhx828: Why not use _.isArray?

kalletlak: I'll change that.

app/scripts/model/dataProxy.js
```diff
@@ -927,7 +927,7 @@ window.DataManagerForIviz = (function($, _) {
                 }
                 // set patient list in studyCasesMap if sample list is
                 // passed in the input
-                if (_studyCasesMap[cancer_study_id].samples !== undefined && _studyCasesMap[cancer_study_id].samples.length > 0) {
```
zhx828: Don't you want to keep the .length > 0?

PR #303, opened 2016-10-06
Code cleanup

kalletlak:
1. Code formatting
2. Fix minor issues when integrated to cbio

### DIFF
--- a/app/scripts/session/sessionServices.js
+++ b/app/scripts/session/sessionServices.js
@@ -1,43 +1,8 @@
-/*
- * Copyright (c) 2015 Memorial Sloan-Kettering Cancer Center.
- *
- * This library is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY, WITHOUT EVEN THE IMPLIED WARRANTY OF MERCHANTABILITY OR FITNESS
- * FOR A PARTICULAR PURPOSE. The software and documentation provided hereunder
- * is on an 'as is' basis, and Memorial Sloan-Kettering Cancer Center has no
- * obligations to provide maintenance, support, updates, enhancements or
- * modifications. In no event shall Memorial Sloan-Kettering Cancer Center be
- * liable to any party for direct, indirect, special, incidental or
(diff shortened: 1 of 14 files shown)


app/scripts/session/sessionServices.js
```diff
@@ -140,22 +107,22 @@ window.vcSession = window.vcSession ? window.vcSession : {};
...
-          });
-          if (virtualCohort_ !== undefined) {
-            iViz.applyVC(_.omit(virtualCohort_,
-              ['created', 'userID', 'virtualCohortID']));
-            jQuery.notify('Imported Virtual Cohort', 'success');
-          } else {
-            jQuery.notify('Error While importing Virtual Cohort', 'error');
-          }
-        });*/
+        /* $.getJSON( vcSession.URL + virtualCohortID, function(response) {
```
zhx828: Should we just remove these code?

app/scripts/views/components/vc/editableRowComponent.js
```diff
@@ -79,11 +79,11 @@
       },
       clickImport: function(_virtualStudy) {
         this.showmodal = false;
-        window.open(window.cbioURL + 'study?cohorts=' + _virtualStudy.virtualCohortID);
+        window.open(window.location.origin + window.location.pathname.substring(0, window.location.pathname.indexOf('/', 2)) + '/study?cohorts=' + _virtualStudy.virtualCohortID);
```
zhx828: Should we make this path as a variable in index.html like cbioURL?

app/scripts/session/sessionServices.js
kalletlak: Sure. Kept it because in future if we want to test visualizing a virtual cohort in iViz.

app/scripts/views/components/vc/editableRowComponent.js
kalletlak: Yes we could. But we need to set this variable wherever we initialized this component. Or we need to declare this variable in global top level js(may be include.js) in cbio so that we can use it for all components for iViz or session management component.

PR #309, opened 2016-10-11
Minor css and js dependency updates

kalletlak:
1. Set default cohort save and manage buttons to false
2. Update session module libraries
3. Remove vue.js from building iviz-vendor js. This is because session module vendor js(which would be added to global library include list) would include this library.


zhx828:
#309 should be merged once we update cbio with session-moulde libraries else it would break existing study-view

yichaoS (approved):

PR #315, opened 2016-10-13
Fix issues in milestone 1.0.3

zhx828:


app/scripts/model/dataProxy.js
```diff
@@ -3,10 +3,11 @@ window.DataManagerForIviz = (function($, _) {
   var content = {};
   // DESC
   var clinicalAttrsPriority = ['CANCER_TYPE', 'CANCER_TYPE_DETAILED',
-    'GENDER', 'AGE', 'sequenced', 'has_cna_data', 'sample_count_patient'];
+    'GENDER', 'SEX', 'AGE', 'sequenced', 'has_cna_data', 'sample_count_patient'];
```
kalletlak: Its gonna be a hard fix to change the code every when ever there is update to this attributes list. We need to find different solution.

kalletlak: May be get them from a json file?

app/resources/attributes.json
```diff
@@ -0,0 +1,38 @@
...
+        "DARWIN_VITAL_STATUS"
+      ]
+    }
+  },
+  
+  "tableAttrs": [
+    "CANCER_TYPE",
+    "CANCER_TYPE_DETAILED"
+  ],
+  "hiddenAttrs": {
```
kalletlak: Could we move this hidden attributes in to study level
Example
"study": {
      "mskimpact": {
        "priorityAttrs":[
          "DARWIN_PATIENT_AGE",
          "OS_STATUS",
          "SAMPLE_TYPE",
          "DARWIN_VITAL_STATUS"
        ],
        "hiddenAttrs": [
          "OS_SURVIVAL",
          "DFS_SURVIVAL",
          "AGE"
        ]
      }
    }

app/index.html
```diff
@@ -68,18 +68,20 @@
...
 
   <!-- build:js scripts/vc-session.js -->
   <script src="_components/vc-session.js"></script>
   <!-- endbuild -->
 
   <!-- build:js scripts/iviz.js -->
+  <script src="_components/Notification.js"></script>
+  <script src="_components/GeneValidator.js"></script>
+  <script src="_components/QueryByGeneUtil.js"></script>
+  <script src="_components/QueryByGeneTextArea.js"></script>
```
kalletlak, line 84: Are we moving these files to iviz.js?

zhx828: I think these files are no longer available in cBioPortal. Either we put them into lib folder, otherwise we need to include in our iViz

app/resources/attributes.json
zhx828: We can change hiddenAttrs to study based. But I think we should separate priority list and hidden list. They are different concepts

kalletlak: If different studies may have different priority list and hidden list then this format might be easy to get those lists.

PR #324, opened 2016-11-10
Minor updates and cleanup

kalletlak:


app/scripts/session/sessionServices.js
```diff
@@ -30,48 +28,47 @@ window.vcSession = window.vcSession ? window.vcSession : {};
...
-          if (_virtualCohort.userID === 'DEFAULT')
-            localStorageEdit_(data.virtualCohort);
+          data: JSON.stringify(_virtualCohort)
+        }).done(function() {
+          if (_virtualCohort.userID === 'DEFAULT') {
+            localStorageEdit_(_virtualCohort);
+          }
         }).fail(function(jqXHR) {
+          // TODO: should we delete the virtual cohort if no record found
+          // in the database? Should we add it into database?
```
yichaoS, line 71: Why delete from local storage and saveSession at the same time? And I think if the session service is not running, or for whatever reason we can't retrieve the VC from session service, we should leave everything as it in the drop down, but when user click on visualize/query just shows an error message

Issue #355, opened 2017-01-13
use gene panel data to calculate percentages in Mutations and CNA table


jjgao:
https://github.com/cBioPortal/cbioportal/issues/2063

PR #361, opened 2017-02-06
Fix issue #355

yichaoS:
https://github.com/cBioPortal/iViz/issues/355
Include gene panel info when calculating alteration rate in Mut & CNA tables

### DIFF
--- a/app/scripts/model/dataProxy.js
+++ b/app/scripts/model/dataProxy.js
@@ -853,6 +853,7 @@ window.DataManagerForIviz = (function($, _) {
       cancerStudyIds: [],
       mutationProfileIdsMap: {},
       cnaProfileIdsMap: {},
+      panelSampleMap: {},
       portalUrl: _portalUrl,
       studyCasesMap: _study_cases_map,
       initialSetup: initialSetup,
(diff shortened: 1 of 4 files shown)


app/scripts/model/dataProxy.js
```diff
@@ -1279,6 +1280,83 @@ window.DataManagerForIviz = (function($, _) {
...
           this.getSampleClinicalData(attribute_ids);
+      },
+      getAllGenePanelSampleIds: window.cbio.util.makeCachedPromiseFunction(
+        function (self, fetch_promise) {
+          var _map = {};
+          $.ajax({
+            type: "GET",
+            url: window.cbioURL + "api-legacy/genepanel/data",
+            contentType: "application/json",
+            data: ["profile_id=genie_public_mutations", "genes="].join("&")
```
kalletlak: is profile_id hard coded?

app/scripts/views/components/dataTable/tableViewTemplate.js
```diff
@@ -196,11 +197,16 @@
       callbacks.submitClick = this.submitClick;
       _self.chartInst = new iViz.view.component.TableView();
       _self.chartInst.setDownloadDataTypes(['tsv']);
-      if (this.isMutatedGeneCna) {
-        $.when(iViz.getTableData(_self.attributes.attr_id))
-          .then(this.processTableData);
+      if (_self.isMutatedGeneCna) {
+        $.when(iViz.getTableData(_self.attributes.attr_id)).then(function(_tableData) {
+          $.when(window.iviz.datamanager.getGenePanelMap(_tableData)).then(function (_genePanelMap) {
```
kalletlak: is _tableData used in getGenePanelMap function?

kalletlak (approved):

app/scripts/model/dataProxy.js
```diff
@@ -1279,6 +1280,94 @@ window.DataManagerForIviz = (function($, _) {
...
+        function (self, fetch_promise) {
+          var _map = {};
+          var asyncAjaxCalls = [];
+          var responses = [];
+          _.each(self.getCancerStudyIds(), function(_studyId) {
+            asyncAjaxCalls.push(
+              $.ajax({
+                url: window.cbioURL + 'api-legacy/genepanel/data',
+                contentType: "application/json",
+                data: ["profile_id=" + _studyId + "_mutations", "genes="].join("&"),
```
kalletlak, line 1294: I think gene list is mandatory for api-legacy/genepanel/data' api. If no gene is passed then the response would be an empty list(object)

yichaoS: I talked to Zack about it. It is better to do it this way instead of passing 600+ genes into it. And the API should be designed to return all gene panels under this study when no genes were specified. I think a legit fix here is asking Zack (who wrote this API) to accept request with no "genes" parameter

PR #368, opened 2017-02-15
Fix minor issues

kalletlak:
- Fix https://github.com/cBioPortal/cbioportal/issues/2096
- Fix selected samples/patients issue
- Fix dc charts groups sync issue

### DIFF
--- a/app/scripts/views/mainTemplate.js
+++ b/app/scripts/views/mainTemplate.js
@@ -197,9 +197,6 @@
         var _resultCounterSelectedCases =
           iViz.util.intersection(_mappedCounterSelectedCases,
             _counterSelectedCasesByFilters);
-        var _resultSelectedCases =
-          iViz.util.idMapping(iViz.getCasesMap(_counterCaseType),
-            _resultCounterSelectedCases);
         var _casesSync = iViz.util.idMapping(iViz.getCasesMap(_counterCaseType),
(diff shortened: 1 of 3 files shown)


app/scripts/views/mainTemplate.js
```diff
@@ -207,7 +209,7 @@
           self_.samplesync = _counterCasesSync;
           if (self_.hasfilters) {
             self_.selectedsamples = _resultCounterSelectedCases;
-            self_.selectedpatients = _resultSelectedCases;
+            self_.selectedpatients = _selectedCasesByFilters;
```
zhx828, line 209: In this case, _resultSelectedCases will never be used. Should we also remove related code?

app/scripts/vueCore.js
```diff
@@ -139,7 +139,7 @@
               self_.checkForDropDownCharts();
               _.every(self_.groups, function(group) {
                 if (group.type === attrData.group_type) {
-                  if (group.attributes.length < 31) {
+                  if (group.attributes.length < 30) {
```
zhx828, line 142: Any reason changed to 30?

kalletlak: dc filtering is not when there are more than one patient or sample groups if it 31. with 30 its working fine

zhx828 (approved):

PR #404, opened 2017-03-07
Fix width issue for long description

kalletlak:


app/scripts/views/components/vc/editableFieldComponent.js
```diff
@@ -6,7 +6,7 @@
     template: '<div v-if="edit"><div v-if="type==\'text\'"><input' +
     ' type="text" v-model="name" placeholder="My Virtual' +
     ' Study"/></div><div v-if="type==\'textarea\'"><textarea rows="4"' +
-    ' cols="50" v-model="name"></textarea></div></div><div' +
+    ' cols="50" v-model="name"></textarea></div></div><div style="white-space:normal !important;"' +
```
yichaoS: Could you update this to "pre-line" instead?

yichaoS: .modal-xlg {
 width: 100% !important;
}
in _modal.scss

PR #408, opened 2017-03-09
Show CANCER_TYPE or CANCER_TYPE_DETAILS first for mskimpact_heme

zhx828:


zhx828:
Waiting approval from JJ.

zhx828:
We want to prioritize CANCER_TYPE and CANCER_TYPE_DETAILS for mskimpact_heme.
JJ and I tried to remove the hard coded restriction on these two attributes and always show them as the first and second chart, but in some studies, the layout is not desirable. We probably need to think about how to do a better layout. But now, just add one more support for mskimpact_heme.

n1zea144:
Thanks.  Responding to your last comment - there are other analysis platforms that will be coming out of the clinical bioinformatics pipeline that will result in other mixed MSK studies in which we will probably want to do the same thing.

yichaoS (approved):

zhx828:
@n1zea144 I've sent a pull request to portal hotfix branch https://github.com/cBioPortal/cbioportal/pull/2192. In the future, could you let me know if we have new studies need similar setting? I need to find a better way handling this issue. :) Thanks!

PR #453, opened 2017-04-07
added back the missing part that is in portal already but not here

yichaoS:


app/scripts/session/sessionUtil.js
```diff
@@ -52,9 +52,19 @@
       _virtualCohort.created = new Date().getTime();
       if (name) {
         _virtualCohort.studyName = name;
+      } else {
+        _virtualCohort.studyName = "Custom Cohort (" + new Date().toISOString().replace(/T/, ' ') + ")";
```
zhx828: Are you planning to move this into a util in another pull request?

zhx828 (approved):

PR #454, opened 2017-04-10
Exclude iViz as dependency of dataProxy; No update on uninitiated tables;

zhx828:


app/scripts/views/components/dataTable/tableView.js
```diff
@@ -100,43 +100,47 @@
...
-                selectedMap_[category].push(caseId);
-                break;
-            }
-            if (isMutatedGeneCna) {
-              _.each(tempData_, function(geneIndex) {
-                if (selectedMap_[geneIndex] === undefined) {
-                  selectedMap_[geneIndex] = {};
-                  if (includeMutationCount) {
-                    selectedMap_[geneIndex].num_muts = 1;
+            if (_.isNumber(caseIndex_)) {
```
kalletlak, line 103: Any specific reason for this check?

zhx828: Originally I thought `caseId` should always have a `caseIndex`, but for some studies (maybe because the data is not curated correctly at server-side), this is incorrect. I saw this issue while testing some other fixes. `caseIndex` may be `undefined`

kalletlak (approved):

yichaoS (approved):

PR #471, opened 2017-04-24
Handle invalid bar chart data

zhx828:
Invalid bar chart will show an error message.
screen shot 2017-04-24 at 6 37 36 pm

### DIFF
--- a/app/scripts/views/components/barChart/barChartTemplate.js
+++ b/app/scripts/views/components/barChart/barChartTemplate.js
@@ -9,14 +9,16 @@
     ':data-number="attributes.priority" @mouseenter="mouseEnter" ' +
     '@mouseleave="mouseLeave">' +
     '<chart-operations :show-log-scale="settings.showLogScale"' +
-    ':show-operations="showOperations" :groupid="attributes.group_id" ' +
+    ':show-operations="showOperations && !failedToInit" :groupid="attributes.group_id" ' +
     ':show-survival-icon.sync="showSurvivalIcon"' +
     ':reset-btn-id="resetBtnId" :chart-ctrl="barChart" ' +
     ':chart-id="chartId" :show-log-scale="showLogScale" ' +
     ':attributes="attributes"' +
     ':filters.sync="attributes.filter"></chart-operations>' +
     '<div class="dc-chart dc-bar-chart" align="center" ' +
-    'style="float:none !important;" id={{chartId}} ></div>' +
+    'style="float:none !important;" id={{chartId}} >' +
+    '<error-handle v-if="failedToInit"></error-handle>' +
+    '</div>' +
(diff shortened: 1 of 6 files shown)


app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -201,9 +202,15 @@
       if (((this.data.max - this.data.min) > 1000) && (this.data.min > 1)) {
         this.settings.showLogScale = true;
       }
-      this.initChart(this.settings.showLogScale);
-      this.updateShowSurvivalIcon();
```
yichaoS: Why line 205 is gone?

zhx828: Thanks! Just added back.

yichaoS (approved):

app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -201,9 +202,16 @@
...
-      this.$dispatch('data-loaded', this.attributes.group_id, this.chartDivId);
+      
+      if (isNaN(this.data.min) || isNaN(this.data.max)) {
```
kalletlak: Is there a possible for this condition. I think all `NaN` are filtered in while finding extremes(`var findExtremeResult = cbio.util.findExtremes(this.data.meta);`). If its a `NaN` shoudn't it throw an error at line 202?

zhx828: Actually, it won't. NaN can be calculated so that's why when `findExtremes ` returns [NaN, NaN, NaN] the line 202 doesn't throw an exception. But you do have a good point, I should check NaN before line 202

app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -188,22 +190,38 @@
 
       this.data.meta = _.map(_.filter(_.pluck(
         iViz.getGroupNdx(this.opts.groupid), this.opts.attrId), function(d) {
+        if (typeof d === 'undefined' || d === 'na' || d === '' ||
```
kalletlak, line 193: could you add a comment of why you are doing such filtering

app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -188,22 +190,38 @@
...
+          d === 'NaN' || d == null) {
+          d = 'NA';
+        }
         return d !== 'NA';
       }), function(d) {
+        var number = parseFloat(d);
+        if (isNaN(number)) {
+          _dataIssue = true;
+        }
```
kalletlak: return could be `number` instead of `parseFloat(d)`

kalletlak (approved):

PR #495, opened 2017-05-10
Allow TSV download for scatter plot

zhx828:


app/scripts/views/components/scatterPlot/scatterPlot.js
```diff
@@ -165,9 +185,39 @@
...
+      } else if (fileType === 'tsv') {
+        initTsvDownloadData();
       }
     };
 
+    function initTsvDownloadData() {
+      var _title = opts_.title || 'Mutation Count vs. CNA';
+      var _data = ['Patient ID', 'Sample ID', 'Mutation Count', 'CNA'];
+
+      if (groups_.length > 1) {
```
kalletlak: Instead of `Group` why not we specify it as `Selected` and values as `0/1 or true/false`

kalletlak: and why not add it by default, I mean for any number of groups?

kalletlak (changes requested):

app/scripts/views/components/scatterPlot/scatterPlot.js
zhx828: I struggled a little bit. If the user just load the page and download the file, should we treatment as all selected?

zhx828: I think about it. I was going to add download for survival plot, and survival has group concept for rainbow survival. Just for consistency, I changed to group in here as well. What do you think?

kalletlak: I think yes!

zhx828: OK, that would make life much easier. lol. I will change that

kalletlak: Ok.. that make sense

kalletlak (approved):

PR #503, opened 2017-05-17
Use priority from database; Improve layout method

### LINKED ISSUE #527: Proper layout charts
Due to the nature of packery.metafizzy.co library, based on the priority, it will automatically fit in charts but may not look good as expected.

Need to do a second sorting and fill in the space with small charts(bar or pie) instead of table

zhx828:
Utilize priorities from database;
Update front-end priority calculation method, see README for details;
Update layout calculation method. Group charts into 4 elements matrix then plot them based on browser width; 
This will close #527


app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,170 @@
+/**
+ * @author Hongxin Zhang on 5/12/17.
+ */
+
+/**
+ * Priorities
+ *
```
yichaoS: priority score?

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,170 @@
...
+ * Rank tier based on the index of priority array. Lower the index, higher
+ * the priority it will be.
+ * Each tier has it's own priorities
+ *
+ * Each clinical attribute has its own priority property with four elements
+ * array. Example: [2, 3, 4, 5]
+ *
+ * Front-end uses tiers to pick clinical attributes. Eventually tiers will be
+ * calculated into a number and each tier is a multiple of 10.
+ * The example above will be calculated to 2345.
```
yichaoS: * The example above will be calculated to 2*1000 + 3* 100 + 4 * 10 + 5 = 2345

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,170 @@
+/**
+ * @author Hongxin Zhang on 5/12/17.
+ */
+
+/**
+ * Priorities
+ *
+ * Priority system has been evolved to tier + priority
```
yichaoS: The priority system is represented with a four elements array, [tier 1 score, tier 2 score, tier 3 score, tier 4 score]. The lower the tier, the more it weights in priority. The weight of each tier is a multiple of 10. For example, a [2, 3, 4, 5] priority array can be calculated into a score 2345 from 2 x 1000 + 3 x 100 + 4 x 10 + 5. The higher the final (numeric) score, the higher priority assigned.

yichaoS (approved):

app/scripts/model/dataProxy.js
```diff
@@ -371,15 +251,21 @@ window.DataManagerForIviz = (function($, _) {
                 _metaObj.filter = [];
                 _metaObj.keys = {};
                 _metaObj.numOfDatum = 0;
-                _metaObj.priority = -1;
-                _metaObj.show = true;
+                if (!_.isArray(_metaObj.priority)) {
+                  iViz.priorityManager
+                    .setClinicalAttrPriority(_metaObj.attr_id, [1, 1, 1, Number(_metaObj.priority)]);
```
kalletlak: why to set the attr priority in priorityManager? why dont we get the priority directly

app/scripts/model/dataProxy.js
```diff
@@ -419,15 +300,21 @@ window.DataManagerForIviz = (function($, _) {
                 _metaObj.filter = [];
                 _metaObj.keys = {};
                 _metaObj.numOfDatum = 0;
-                _metaObj.show = true;
-                _metaObj.priority = -1;
+                if (!_.isArray(_metaObj.priority)) {
+                  iViz.priorityManager
+                    .setClinicalAttrPriority(_metaObj.attr_id, [1, 1, 1, Number(_metaObj.priority)]);
```
kalletlak: same here

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,170 @@
...
+      return number;
+    };
+
+    content.comparePriorities = function(aP, bP, asc) {
+      var _a = this.calcPriorityNumber(aP);
+      var _b = this.calcPriorityNumber(bP);
+      return asc ? (_a - _b) : (_b - _a);
+    };
+
+    content.compareClinicalAttr = function(aAttr, bAttr, asc) {
```
kalletlak: are we using this the function anywhere?

app/scripts/model/dataProxy.js
zhx828: PriorityManager needs to remember each clinical attribute priority in order to calculate the average score for combination charts

app/scripts/controller/priorityManager.js
zhx828: It's legacy code. I can remove it.

app/scripts/model/dataProxy.js
kalletlak: I feel like priority manager shouldn't save the priority for each and every attribute.. why dont we pass array of attributes priorities to `getSecondTierPriority`

zhx828: But you still need to loop through all clinical attributes to get these priorities?

kalletlak: I think its ok to loop thought attributes. This way portal manager should would be much cleaner and easy to maintain.

zhx828:
@kalletlak @yichaoS I updated the pull request. I still suggest managing the priority inside priority manager. Otherwise I need to lots of if/else check for some special clinical attributes like CANCER_TYPE

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,159 @@
...
+     * Get forth tier priority by clinical attribute.
+     * @param{string} id Clinical attribute ID.
+     * @return {number}
+     */
+    function getForthTierPriority(id) {
+      return clinicalAttrsPriority.hasOwnProperty(id)
+        ? clinicalAttrsPriority[id][3] : 1;
+    }
+
+    function getDefaultPriority() {
```
kalletlak: Could we make this as a variable instead of function

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Convert priority array to a number
+     * @param {array} priority
+     * @return {number}
+     */
+    content.calcPriorityNumber = function(priority) {
+      var number = 0;
+      if (_.isArray(priority)) {
+        var _arrL = priority.length - 1;
+        for (var i = _arrL; i >= 0; i--) {
```
kalletlak: May be `parseInt(priority.join(""))`

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,166 @@
...
+        ? clinicalAttrsPriority[id][3] : 1;
+    }
+
+    /**
+     * Convert priority array to a number
+     * @param {array} priority
+     * @return {number}
+     */
+    content.calcPriorityNumber = function(priority) {
+      var number = 0;
```
kalletlak: may be in one line `(_.isArray(priority)) ? parseInt(priority.join('')) : 0`

app/scripts/controller/priorityManager.js
```diff
@@ -0,0 +1,166 @@
...
+ */
+
+/**
+ * Priorities
+ *
+ * The priority system is represented with a four elements array,
+ * [tier 1 score, tier 2 score, tier 3 score, tier 4 score].
+ * The lower the tier, the more it weights in priority.
+ * The weight of each tier is a multiple of 10.
+ * For example, a [2, 3, 4, 5] priority array can be calculated into
```
kalletlak: question. instead of considering priority as a array of numbers `[2, 3, 4, 5]` and later converting it to a single number `2345`, why don't you use the final number directly i.e. `2345`

app/scripts/controller/util.js
```diff
@@ -594,8 +356,306 @@
...
+        if (tempArr_[obj_] === undefined) {
+          tempArr_[obj_] = true;
+        }
+      });
+      return Object.keys(tempArr_);
+    };
+
+    content.isRangeFilter = function(filterObj) {
+      if (filterObj.filterType !== undefined) {
+        if (filterObj.filterType === 'RangedFilter') {
```
kalletlak: combine `if` statements

app/scripts/model/dataProxy.js
```diff
@@ -388,16 +267,25 @@ window.DataManagerForIviz = (function($, _) {
                 _metaObj.filter = [];
                 _metaObj.keys = {};
                 _metaObj.numOfDatum = 0;
-                _metaObj.priority = -1;
-                _metaObj.show = true;
+                if (!_.isArray(_metaObj.priority)) {
+                  iViz.priorityManager
+                    .setClinicalAttrPriority(_metaObj.attr_id, [0, 0, 0, Number(_metaObj.priority)]);
```
kalletlak: what if `_metaObj.priority` greater than 9?

app/styles/partials/_jsonVars.scss
```diff
@@ -3,4 +3,8 @@ $grid-w-2:400px;
 $grid-h-1:170px;
 $grid-h-2:350px;
 $numOfSurvivalCurveLimit:20;
-$dc:("transitionDuration":400);
\ No newline at end of file
+$numOfChartsLimit:30;
+$pie2TableLimit:20;
+$dc:("transitionDuration":400);
+$tableAttrs:("CANCER_TYPE","CANCER_TYPE_DETAILED");
+$priority:("CANCER_TYPE":(3,0,0,0),"CANCER_TYPE_DETAILED":(2,0,0,0),"OS_SURVIVAL":(0,4,0,0),"DFS_SURVIVAL":(0,3,0,0),"MUT_CNT_VS_CNA":(0,2,0,0),"GENDER":(0,0,0,9),"SEX":(0,0,0,9),"AGE":(0,0,0,8),"mutated_genes":(0,0,9,0),"cna_details":(0,0,8,0),"study_id":(0,0,7,0),sequenced:(0,0,6,0),"has_cna_data":(0,0,5,0),"sample_count_patient":(0,0,4,0),"mutation_count":(0,0,3,0),"copy_number_alterations":(0,0,2,0));
```
kalletlak: are you using these anywhere or are you getting them from `config.json` file?

app/scripts/controller/priorityManager.js
zhx828: We could, thought in this way the logic is clearer.

app/scripts/controller/priorityManager.js
zhx828: Will do

app/scripts/controller/util.js
zhx828: Will do

app/scripts/model/dataProxy.js
zhx828: We take that. It will be upgraded to the upper level.

app/styles/partials/_jsonVars.scss
zhx828: We don't use them in scss files actually. This is just the set up reading variables from config.json and convert them into scss variables automatically. It may create some confusion in the future. Any suggestions?

app/scripts/model/dataProxy.js
kalletlak: :thumbsup:

app/scripts/controller/priorityManager.js
kalletlak: but it adds additional complexity.

app/styles/partials/_jsonVars.scss
kalletlak: nothing in my mind for now

app/scripts/controller/priorityManager.js
zhx828: Yeah, agree. Don't know the better the solution yet.

Issue #490, opened 2017-05-03
Exclude NA in survival plot


zhx828:
Give option to include/exclude NA samples in the tooltip
![screen shot 2017-05-03 at 5 48 00 pm](https://cloud.githubusercontent.com/assets/5400599/25683115/bb0865fe-3028-11e7-8bdf-b873ba975370.png)

For Mutated Genes and CNA tables, only include sequenced samples

PR #518, opened 2017-05-31
Exclude NA cases for survival plot by default on selected clinical attributes

zhx828:
Related to https://github.com/cBioPortal/iViz/issues/490
Heroku instance: https://survival-exclude-na.herokuapp.com/

New interface:
screen shot 2017-06-15 at 5 18 28 pm

### DIFF
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -67,7 +67,7 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
         attrData.group_type = 'patient';
         if (chartsCount < 20 && patientChartsCount < 10) {
           if (attrData.show) {
-            attrData.group_id = vm_.groupCount;
+            attrData.group_id = vm_.groupCount.toString();
             groupAttrs.push(attrData);
             chartsCount++;
             patientChartsCount++;
@@ -82,7 +82,7 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
       });
       groups.push({
         type: 'patient',
-        id: vm_.groupCount,
+        id: vm_.groupCount.toString(),
         selectedcases: [],
         hasfilters: false,
         attributes: groupAttrs});
@@ -93,7 +93,7 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
         attrData.group_type = 'sample';
         if (chartsCount < 20) {
           if (attrData.show) {
-            attrData.group_id = vm_.groupCount;
+            attrData.group_id = vm_.groupCount.toString();
             groupAttrs.push(attrData);
             chartsCount++;
           }
@@ -105,7 +105,7 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
 
       groups.push({
         type: 'sample',
-        id: vm_.groupCount,
+        id: vm_.groupCount.toString(),
         selectedcases: [],
         hasfilters: false,
         attributes: groupAttrs});
--- a/app/scripts/views/components/survivalChart/template.js
+++ b/app/scripts/views/components/survivalChart/template.js
@@ -27,12 +27,9 @@
     },
     data: function() {
       return {
-        chartDivId:
-          iViz.util.getDefaultDomId('chartDivId', this.attributes.attr_id),
-        resetBtnId:
-          iViz.util.getDefaultDomId('resetBtnId', this.attributes.attr_id),
-        chartId:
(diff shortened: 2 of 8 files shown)


app/scripts/views/components/survivalChart/template.js
```diff
@@ -50,17 +49,94 @@
...
         var attrId = _type === 'patient' ? 'patient_uid' : 'sample_uid';
         var _selectedCases = [];
-        var _allCases = iViz.getCaseUIDs(_type);
+        var _allCases = iViz.getCaseUIDs(_type).sort();
         var groups = [];
 
-        if (hasFilters) {
-          _selectedCases =
-            _.pluck(this.invisibleDimension.top(Infinity), attrId);
+        if (_.isBoolean(excludeNa)) {
```
yichaoS: Why need to use "isBoolean" here?

app/scripts/views/components/survivalChart/template.js
```diff
@@ -50,17 +49,94 @@
...
+          var filteredClinicalAttrs = _.flatten(_.map(this.$root.groups, function(group) {
+            return _.pluck(_.filter(group.attributes, function(attr) {
+              return attr.filter.length > 0;
+            }), 'attr_id');
+          }));
+          if (this.excludeNa) {
+            var data_ = iViz.getGroupNdx(this.attributes.group_id);
+            var _nonNaCases = [];
+            _.each(data_, function(data) {
+              var _intersection = _.intersection(Object.keys(data), filteredClinicalAttrs);
```
yichaoS: I remember _.intersection has been causing performance issues? I see we are using iViz.util.intersection everywhere else.

zhx828: Yeah, but this intersection should have a very minimium impact. I have used iViz.uti.intersection for other places

app/scripts/views/components/survivalChart/template.js
zhx828: When the moment `updatePlot` is called, two parameters may not both available. But, yeah, two places where use this function are passing both parameters correctly anyway... I can remove the check.

app/scripts/views/components/survivalChart/template.js
```diff
@@ -50,17 +49,94 @@
...
+              var _isNa = false;
+              _.some(_intersection, function(attr) {
+                if (data[attr].toString().toLowerCase() === 'na') {
+                  _isNa = true;
+                  return true;
+                }
+              });
+              if (!_isNa) {
+                _nonNaCases.push(data[attrId]);
```
yichaoS: Can move this if into _.some therefore no need to create _isNa?

zhx828: Ah, good point. I will update. Thanks!

yichaoS (approved):

app/scripts/main.js
```diff
@@ -67,7 +67,7 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
         attrData.group_type = 'patient';
         if (chartsCount < 20 && patientChartsCount < 10) {
           if (attrData.show) {
-            attrData.group_id = vm_.groupCount;
+            attrData.group_id = vm_.groupCount.toString();
```
kalletlak, line 70: any specific reason to convert to string

yichaoS (approved):

app/scripts/views/components/survivalChart/template.js
```diff
@@ -122,21 +94,141 @@
...
+        // Display name may be changed due to the rainbow survival
+        this.displayName = this.attributes.display_name;
+
+        this.chartInst.update(
+          groups ? groups : this.groups, this.chartId, this.attributes.attr_id);
+        this.checkDownloadableStatus();
+        this.showLoad = false;
+        this.updateQtipContent();
+      },
+      updatePlotGroups: function(hasFilters) {
```
kalletlak, line 125: Since this function going to mutate the groups data, better to move this to a parent object i.e may be to vueCore.js or main.js

app/scripts/main.js
zhx828: I notice group_id is not always string or number throughout the project which cases some issues

app/scripts/views/components/survivalChart/template.js
zhx828: This is not the groups in the main.js/vueCore.js but the survival groups. It only exists under survival

app/scripts/main.js
kalletlak: ok

app/scripts/views/components/survivalChart/template.js
kalletlak: ok

kalletlak (approved):

PR #520, opened 2017-06-06
Create Scientific log for small numbers.

victoria34:
Group small numbers (most of them < 0.001) by their exponents. Use scientific notation to format ticks.

### DIFF
--- a/app/_components/cbio-util.js
+++ b/app/_components/cbio-util.js
@@ -672,7 +672,6 @@ cbio.util = (function() {
     return def.promise();
   }
 
-  //Jing's function: get exponent for 0<data<0.001
 
   function getDecimalExponents(data){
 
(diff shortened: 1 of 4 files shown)


app/_components/cbio-util.js
```diff
@@ -666,6 +672,42 @@ cbio.util = (function() {
...
+      }
+    });
+
+    // Then sort
+    values.sort(function(a, b) {
+      return a - b;
+    });
+    
+    var minZeros = 0;
+    while (values[0] < 1) {
```
zhx828: Need to make sure `values` has content in it.

app/index.html
```diff
@@ -209,11 +209,11 @@
 
     window.cbioURL = 'http://www.cbioportal.org/';
     window.cbioResourceURL = 'resources/';
-    window.mutationProfileId = 'hnsc_tcga_mutations';
```
zhx828: Don't change the example study here.

app/scripts/views/components/barChart/barChart.js
```diff
@@ -148,6 +177,14 @@
       } else {
         _returnValue = v;
       }
+
+      if (data_.smallDataFlag) {
```
zhx828, line 181: This data needs to be moved into `barChartTemplate.js` at line 218.

zhx828 (changes requested):

PR #522, opened 2017-06-07
Scientific log for small numbers

victoria34:
Add minExponent and maxExponent. Check if data of getDecimalExponents is empty.

### DIFF
--- a/app/_components/cbio-util.js
+++ b/app/_components/cbio-util.js
@@ -672,7 +672,6 @@ cbio.util = (function() {
     return def.promise();
   }
 
-  //Jing's function: get exponent for 0<data<0.001
 
   function getDecimalExponents(data){
 
@@ -708,16 +707,6 @@ cbio.util = (function() {
     }
     return expoents;
   }
-  
-  
-  function getScientificNumber(data){
-
-    var maxValue = 1 / Math.pow(10, Number(data[0]));
-    var minValue = 1 / Math.pow(10, Number(data[1]));
-
-    return [minValue, maxValue];
-    
-  }
 
   return {
     toPrecision: toPrecision,
@@ -747,8 +736,7 @@ cbio.util = (function() {
     deepCopyObject: deepCopyObject,
     makeCachedPromiseFunction: makeCachedPromiseFunction,
     getDatahubStudiesList: getDatahubStudiesList,
-    getDecimalExponents: getDecimalExponents,
-    getScientificNumber:getScientificNumber
+    getDecimalExponents: getDecimalExponents
   };
 
 })();
--- a/app/_components/cbio-util.js
+++ b/app/_components/cbio-util.js
@@ -598,14 +598,18 @@ cbio.util = (function() {
     // Likewise for q3. 
     var q3 = values[(Math.ceil((values.length * (3 / 4))) > values.length - 1 ? values.length - 1 : Math.ceil((values.length * (3 / 4))))];
     var iqr = q3 - q1;
+
     if (values[Math.ceil((values.length * (1 / 2)))] < 0.001) {
       smallDataFlag = true;
     }
     // Then find min and max values
     var maxValue, minValue;
-    if (q3 < 1) {
-      maxValue = Number((q3 + iqr * 1.5).toFixed(2));
-      minValue = Number((q1 - iqr * 1.5).toFixed(2));
+    if (0.001 <= q3 && q3 < 1) {
+      maxValue = Number((q3 + iqr * 1.5).toFixed(3));
+      minValue = Number((q1 - iqr * 1.5).toFixed(3));
+    } else if(q3 < 0.001){// get IQR for very small number(<0.001)
+      maxValue = Number((q3 + iqr * 1.5));
+      minValue = Number((q1 - iqr * 1.5));
     } else {
       maxValue = Math.ceil(q3 + iqr * 1.5);
       minValue = Math.floor(q1 - iqr * 1.5);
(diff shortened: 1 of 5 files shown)


app/_components/cbio-util.js
```diff
@@ -666,6 +672,47 @@ cbio.util = (function() {
     return def.promise();
   }
 
+
+  function getDecimalExponents(data){
+
+    // Copy the values, rather than operating on references to existing values
+    if (data === null || data.length < 1) {//if data is null or empty, return data
```
zhx828: You should use `_.isArray(data)` instead of `data === null`

app/_components/cbio-util.js
```diff
@@ -666,6 +672,47 @@ cbio.util = (function() {
...
+    
+    var values = [];
+    _.each(data, function(item) {
+      if (!isNaN(item)) {
+        values.push(Number(item));
+      }
+    });
+
+    // Then sort
+    values.sort(function(a, b) {
```
zhx828, line 694: The default sorting is a-b. You should use `values.sort()` directly

app/_components/cbio-util.js
```diff
@@ -666,6 +672,47 @@ cbio.util = (function() {
...
+      return a - b;
+    });
+    
+    var minZeros = 0, maxZeros = 0;
+      
+    while (values[0] < 1) {
```
zhx828: Like I said before, `values` is not guaranteed to have any elements

app/scripts/views/components/barChart/barChart.js
```diff
@@ -89,14 +105,19 @@
       if (logScale) {
         chartInst_.x(d3.scale.log().nice()
           .domain([0.7, opts_.maxDomain]));
+      } else if (data_.smallDataFlag) {
```
zhx828: Can't this part be merged with line 115-120?

app/scripts/views/components/barChart/barChart.js
```diff
@@ -131,8 +154,14 @@
           _returnValue = 'NA';
         }
       } else if (v === opts_.xDomain[0]) {
+        if (data_.smallDataFlag) {
+          return '<=' + e(opts_.xDomain[1]);
+        }
         return '<=' + opts_.xDomain[1];
       } else if (v === opts_.xDomain[opts_.xDomain.length - 2]) {
+        if (data_.smallDataFlag) {
```
zhx828: This part should be merged with line 165

app/scripts/views/components/barChart/barChart.js
```diff
@@ -363,6 +413,18 @@
         } else if (max < 100 &&
           max > 10) {
           config.divider = 2;
+        } else if (data.smallDataFlag) {
+          if (exponentRange > 20) {
+            config.divider = 8;
```
zhx828: Is the divider 8 sufficient enough if exponentRange is 100?

zhx828 (changes requested):

app/_components/cbio-util.js
victoria34: The default sort() should return values from small numbers to large values. But in here, it works conversely, so I have to write 'a-b'.

app/scripts/views/components/barChart/barChart.js
victoria34: I'm afraid not. Line 115-120 uses linear() scale, but scientific numbers use log() scale.

app/_components/cbio-util.js
victoria34: Good Idea. Thanks!

app/scripts/views/components/barChart/barChart.js
victoria34: var e = d3.format('.1e');// convert small data to scientific notation format
For scientific numbers, they should be covert to scientific format by e(). If I merged them, normal numbers will also show in scientific format.

app/_components/cbio-util.js
victoria34: You're right. I will use a for loop to make sure min and max values are valid numbers.

app/_components/cbio-util.js
zhx828: a-b is from small to large values

app/scripts/views/components/barChart/barChart.js
zhx828: sorry 106-107

app/scripts/views/components/barChart/barChart.js
zhx828: You still need the if condition, but parts `return '>' + ` and `opts_.xDomain[opts_.xDomain.length - 3]` are the same

app/scripts/views/components/barChart/barChart.js
victoria34: if (data.smallDataFlag) {
          var numberOfTickValues = 4;
          config.divider = Math.round(exponentRange / numberOfTickValues);
}
Now config.divider is calculated automatically.

app/_components/cbio-util.js
victoria34: I didn't explain clearly. Values.sort() in here return data based on kind of alphabetical order, so I have to mention I want to sort data in ascending order.

app/scripts/views/components/barChart/barChart.js
victoria34: if (logScale || data_.smallDataFlag) {
        chartInst_.x(d3.scale.log().nice()
          .domain([opts_.minDomain, opts_.maxDomain]));
 }
Sure, I have merged them.

app/scripts/views/components/barChart/barChart.js
victoria34: formattedValue = opts_.xDomain[1];
        if (data_.smallDataFlag) {
          return '<=' + e(formattedValue);
        }
        return '<=' + formattedValue;

PR #537, opened 2017-06-23
Make scatter async

victoria34:


app/scripts/main.js
```diff
@@ -392,6 +401,78 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+                    _mutationCountData[_sampleDatum.study_id][_sampleDatum.sample_id] === null) {
+                    if (_sequencedCaseUIdsMap[_sampleDatum.sample_uid] === undefined) {
+                      _sampleDatum.mutation_count = 'NA';
+                    } else {
+                      _sampleDatum.mutation_count = 0;
+                    }
+                  } else {
+                    _sampleDatum.mutation_count = _mutationCountData[_sampleDatum.study_id][_sampleDatum.sample_id];
+                  }
+                  mutationCountMap_[_sampleDatum.sample_id] = _sampleDatum.mutation_count;
```
zhx828: If you need to cache mutationCount, then you should use sample_uid instead of sample_id

victoria34: I will change it. Thanks!

app/scripts/main.js
```diff
@@ -392,6 +402,84 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+      var self = this;
+      var data = self.getGroupNdx(groupId);
+
+      $.when(window.iviz.datamanager.getStudyToSampleToPatientdMap()).then(function(_studyToSampleToPatientMap) {
+        $.when(window.iviz.datamanager.getCaseLists()).then(function(_caseLists) {
+          var _sequencedCaseUIDs = [];
+          var _allCaseUIDs = [];
+          var _sequencedCaseUIdsMap = {};
+          
+          $.each(_caseLists, function(studyId, caseList) {
```
zhx828: Lines 416-431 are duplicate code. Please consolidate them.

app/scripts/main.js
```diff
@@ -392,6 +402,84 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
+            _sequencedCaseUIdsMap[_sampleUid] = _sampleUid;
+          });
+
+          $.when(window.iviz.datamanager.getCnaFractionData(),
+            window.iviz.datamanager.getMutationCount())
+            .then(function(_cnaFractionData, _mutationCountData) {
+              var _hasCNAFractionData = _.keys(_cnaFractionData).length > 0;
+              var _hasMutationCountData = _.keys(_mutationCountData).length > 0;
+
+              if ((!_hasCNAFractionData || !_hasMutationCountData) && firstScatterShow) {
```
zhx828: What's this `firstScatterShow ` for?

app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -186,7 +188,15 @@
         width: this.settings.width,
         height: this.settings.height
       });
-
+      groupNdxData = iViz.getGroupNdx(this.opts.groupid);
+      
+      if(this.opts.attrId === 'mutation_count') {
```
zhx828: mutation count data should be added right after the data is ready. It should follow similar `addChart` follow in vueCore.js

zhx828 (changes requested):

app/scripts/main.js
victoria34: firstScatterShow is a flag. When the page is loaded, firstScatterShow is true so the code will remove empty chart and it will be changed to false. When a user wants to add an empty chart, it won't run into removing empty chart part since firstScatterShow is false even if (!_hasCNAFractionData || !_hasMutationCountData) is always true.

app/scripts/main.js
victoria34: You're right. Thanks! I removed them and made _sequencedCaseUIdsMap as a parameter sent to scatterPlotTemplate.js

app/scripts/views/components/barChart/barChartTemplate.js
victoria34: I add mutation_count to data_.groups.sample.data[_sampleDatum.sample_uid].mutation_count

app/scripts/main.js
```diff
@@ -406,78 +403,60 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
-
-              if ((!_hasCNAFractionData || !_hasMutationCountData) && firstScatterShow) {
-                charts['MUT_CNT_VS_CNA'].show = false;
-                firstScatterShow = false;
-                _self.$dispatch('remove-chart', 'MUT_CNT_VS_CNA', groupId);//rearrange layout
-              }
+          // firstScatterShow is a global variable. If web page is just initialized, firstScatterShow is true and 
+          // it turns to false when cna.json or mutation.json returns null. When people want to add the empty 
+          // scatter chart, firstScatterChart(false) ensures the "remove empty chart" module won't be executed.
+          if ((!_hasCNAFractionData || !_hasMutationCountData) && firstScatterShow) { // remove empty scatter chart when web page just loaded
```
zhx828: From my point of view,  `firstScatterShow ` is not really necessary here. The dispatch event could be sent. `remove-chart' ` event needs to handle whether the chart is exist

victoria34: Sorry, I didn't explain clearly. I don't want the dispatch event to be sent when people just add a chart. It should be only triggered when the page is loading. Otherwise, it will always remove the chart instead of generating an empty scatterplot.

app/scripts/model/dataProxy.js
```diff
@@ -569,95 +577,54 @@ window.DataManagerForIviz = (function($, _) {
                   has_attr_data: _hasSampleAttrData
                 }
               };
+              
+              // add Mutation count vs. CNA fraction
+             //if (_hasSampleAttrData.mutation_count !== undefined && _hasSampleAttrData.cna_fraction !== undefined) {
```
zhx828: Let's remove this comment and also line 601

app/scripts/views/components/scatterPlot/scatterPlotTemplate.js
```diff
@@ -138,26 +144,49 @@
...
+            _self.chartInst.setDownloadDataTypes(['pdf', 'svg', 'tsv']);
+            _self.chartInst.init(_scatterData, _opts);
+
+            _self.dataLoaded = true;
+            var _selectedCases =
+              _.pluck(_self.invisibleDimension.top(Infinity), attrId);
+            if (_self.hasFilters) {
+              _self.chartInst.update(_selectedCases);
+            } else {
+              _self.chartInst.update([]);
```
zhx828: If there is no filters, the update function shouldn't be called

zhx828 (changes requested):

app/scripts/model/dataProxy.js
victoria34: Sure.

app/scripts/views/components/scatterPlot/scatterPlotTemplate.js
victoria34: I will remove else statement.

zhx828 (approved):

PR #545, opened 2017-07-06
Fix exclude-na issue when patient/sample doesn't have data on selected attrs

zhx828:


kalletlak (approved):

app/scripts/controller/util.js
```diff
@@ -685,6 +685,28 @@
...
+    content.strIsNa = function(str, includeEmptyStr) {
+      var status = false;
+      includeEmptyStr = _.isBoolean(includeEmptyStr) ? includeEmptyStr : false;
+      if (_.isString(str)) {
+        if (['na', 'nan', 'n/a'].indexOf(str.toLowerCase()) > -1) {
+          status = true;
+        } else if (includeEmptyStr && !str) {
+          status = true;
+        }
+      } else if (typeof str === 'undefined' || str === null) {
```
kalletlak: not sure if `if` statement is required here

app/scripts/controller/util.js
```diff
@@ -685,6 +685,28 @@
...
+     * @param includeEmptyStr Whether empty string should be treated as NA
+     * @returns {boolean}
+     */
+    content.strIsNa = function(str, includeEmptyStr) {
+      var status = false;
+      includeEmptyStr = _.isBoolean(includeEmptyStr) ? includeEmptyStr : false;
+      if (_.isString(str)) {
+        if (['na', 'nan', 'n/a'].indexOf(str.toLowerCase()) > -1) {
+          status = true;
+        } else if (includeEmptyStr && !str) {
```
kalletlak: Is there a possibility for this if statement to pass?

zhx828: Empty string

app/scripts/controller/util.js
zhx828: We need it. The first if only handles String, this one only handles undefined and null. We don't deal with Number, Object or other types

app/scripts/controller/util.js
kalletlak: instead of else if, why not or condition in above if statement

app/scripts/controller/util.js
kalletlak: isn't the input to this function is String?

zhx828: Ideally, but could be other types

app/scripts/controller/util.js
zhx828: Cool

PR #547, opened 2017-07-07
Only hide specific icons if chart is not initialized

zhx828:


kalletlak:
is it just for bar-chart?

kalletlak (approved):

zhx828:
@kalletlak yup. Only bar chart has a strict requirement on the data.

PR #566, opened 2017-07-24
Fix the bug: exclude NA does not work in combination chart

zhx828:
Close issue https://github.com/cBioPortal/cbioportal/issues/2800;
Do not highlight the first survival node in each curve which is created
as a fake node;
Don’t change the input data array directly;

### DIFF
--- a/app/scripts/views/components/survivalChart/template.js
+++ b/app/scripts/views/components/survivalChart/template.js
@@ -144,10 +144,16 @@
                 nonNaCases: []
               };
             }
-            filteredClinicalAttrs[group.id].attrs =
-              _.pluck(_.filter(group.attributes, function(attr) {
-                return attr.filter.length > 0;
-              }), 'attr_id');
+            filteredClinicalAttrs[group.id].attrs = [];
+            
+            // Loop through attrList instead of only using attr_id
+            // Combination chart has its own attr_id, but the clinical data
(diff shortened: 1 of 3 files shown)


app/scripts/views/components/survivalChart/template.js
```diff
@@ -144,10 +144,16 @@
...
-              }), 'attr_id');
+            filteredClinicalAttrs[group.id].attrs = [];
+            
+            // Loop through attrList instead of only using attr_id
+            // Combination chart has its own attr_id, but the clinical data
+            // it's using are listed under attrList
+            _.each(_.filter(group.attributes, function(attr) {
+              return attr.filter.length > 0;
+            }), function(item) {
+              filteredClinicalAttrs[group.id].attrs.push(_.pick(item, 'attr_id', 'attrList'));
```
kalletlak, line 155: not sure if you can pluck two attributes from object

kalletlak: I think it just plucks the first attribute

kalletlak (changes requested):

app/scripts/views/components/survivalChart/template.js
kalletlak: Sorry! my bad... its pick

kalletlak (approved):

PR #573, opened 2017-07-27
Do not reset bar chart color if bar is not initialized

zhx828:


zhx828:
This will solve issue https://github.com/cBioPortal/cbioportal/issues/2869

victoria34 (approved):

PR #577, opened 2017-08-01
Make mutation_count barchart async

victoria34:
Make mutation_count barchart async
Add hasNA for "NA" value and modify gutter when noGrouping is true
Change breadcrumb format


app/scripts/main.js
```diff
@@ -482,6 +482,16 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
             }
           });
+
+          // Load barchart after data is ready
+          charts['mutation_count'].dataLoaded = true;
+          if (!_hasMutationCountData) {
+            charts['mutation_count'].emptyChart = true;
+            if (charts['mutation_count'].addChartBy === 'default') {
+              charts['mutation_count'].show = false;
+              _self.$dispatch('remove-chart', charts['mutation_count'].attr_id, charts['mutation_count'].group_id);//rearrange layou
```
zhx828, line 494: type should be `layout`

zhx828 (changes requested):

PR #579, opened 2017-08-08
Async mutation count bar chart & breadcrumb format

victoria34:
Make mutation_count bar chart async.
Add hasNA for "NA" value and modify gutter when noGrouping is true. 
Create empty chart when mutation_count data is null. 
Reset tick format about NA. 
Change breadcrumb format. 
Add getMutationCountData() and avoid duplicate charts in updateDataObject().

### DIFF
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -447,8 +447,12 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
         self.getMutationCountData(_self))
         .then(function(_cnaFractionData, _mutationCountResult) {
           var _hasCNAFractionData = _.keys(_cnaFractionData).length > 0;
-          var _hasMutationCountData = _mutationCountResult[1];
+          var _hasMutationCountData = false;
           var groupId = _self.attributes.group_id;
+
+          if (_mutationCountResult[1]) {
+            _hasMutationCountData = _mutationCountResult[1];
+          }
           data = self.getGroupNdx(groupId);
 
           _.each(data, function(_sampleDatum) {
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -444,29 +444,14 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
       var data = {};
       
       $.when(window.iviz.datamanager.getCnaFractionData(),
-        window.iviz.datamanager.getMutationCount())
-        .then(function(_cnaFractionData, _mutationCountData) {
+        self.getMutationCountData(_self))
+        .then(function(_cnaFractionData, _mutationCountResult) {
           var _hasCNAFractionData = _.keys(_cnaFractionData).length > 0;
-          var _hasMutationCountData = _.keys(_mutationCountData).length > 0;
+          var _hasMutationCountData = _mutationCountResult[1];
           var groupId = _self.attributes.group_id;
           data = self.getGroupNdx(groupId);
-          
-          _.each(data, function(_sampleDatum) {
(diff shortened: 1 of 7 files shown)


app/scripts/main.js
```diff
@@ -441,13 +444,44 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
       var data = {};
       
       $.when(window.iviz.datamanager.getCnaFractionData(),
-        window.iviz.datamanager.getMutationCount())
-        .then(function(_cnaFractionData, _mutationCountData) {
+        self.getMutationCountData(_self))
+        .then(function(_cnaFractionData, _mutationCountResult) {
           var _hasCNAFractionData = _.keys(_cnaFractionData).length > 0;
+          var _hasMutationCountData = _mutationCountResult[1];
```
zhx828: `_hasMutationCountData` should have a default value? In case, _mutationCountResult is not array or for some reason `self. getMutationCountData ` failed?

zhx828 (changes requested):

app/scripts/main.js
victoria34: Good idea! I will change it. Thanks!

app/scripts/main.js
```diff
@@ -447,8 +447,12 @@ var iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
         self.getMutationCountData(_self))
         .then(function(_cnaFractionData, _mutationCountResult) {
           var _hasCNAFractionData = _.keys(_cnaFractionData).length > 0;
-          var _hasMutationCountData = _mutationCountResult[1];
+          var _hasMutationCountData = false;
           var groupId = _self.attributes.group_id;
+
+          if (_mutationCountResult[1]) {
```
zhx828: If `_mutationCountResult ` is not array, it will throw error here

app/scripts/views/components/barChart/barChartTemplate.js
```diff
@@ -17,13 +17,14 @@
     ':chart-id="chartId" :show-log-scale="showLogScale" ' +
     ':attributes="attributes"' +
     ':filters.sync="attributes.filter"></chart-operations>' +
+    '<div class="dc-chart dc-bar-chart" align="center" ' +
+    'style="float:none !important;" id={{chartId}} >' +
```
zhx828: Why introduce a `important`?

app/scripts/views/components/scatterPlot/scatterPlotTemplate.js
```diff
@@ -147,9 +153,14 @@
       
       $.when(iViz.getScatterData(_self))
         .then(function(_scatterData, _hasCnaFractionData, _hasMutationCountData) {
-          if ((!_hasCnaFractionData || !_hasMutationCountData) && _self.attributes.addChartBy === 'default') {
-            _self.attributes.show = false;
-            _self.$dispatch('remove-chart', _self.attributes.attr_id,  _self.attributes.group_id);//rearrange layout
+          if (!_hasCnaFractionData || !_hasMutationCountData) {
+            if (_self.attributes.addChartBy === 'default') {
+              _self.attributes.show = false;
```
zhx828, line 165: Even when added by default, the errorMessage.noData and failedToInit should be set to true as well?

zhx828 (changes requested):

app/scripts/views/components/scatterPlot/scatterPlotTemplate.js
victoria34: Changed

zhx828 (approved):

PR #586, opened 2017-08-15
Reframe Breadcrumb rangeFilter()

victoria34:
Show selected range based on its ticks rather than inner bars.

### DIFF
--- a/app/scripts/views/components/barChart/barChart.js
+++ b/app/scripts/views/components/barChart/barChart.js
@@ -306,27 +306,29 @@
       var hasGreaterOutlier = false;
       var hasSmallerOutlier = false;
       var startNumIndex = 0;
-      var endNumIndex = opts_.xDomain.length - 1;
+      var endNumIndex = opts_.xDomain.length >= 1 ? opts_.xDomain.length - 1 : 0;
 
       tempFilters_[0] = '';
       tempFilters_[1] = '';
       
       // set start and end indexes for number bars
-      if (data_.noGrouping) {
-        if (data_.hasNA) {
-          endNumIndex = opts_.xDomain.length - 2;
-        } 
-      } else {
-        if (logScaleChecked) {
-          if (data_.hasNA) { // has 'NA' tick
-            endNumIndex = opts_.xDomain.length - 2;
-          } 
+      if (opts_.xDomain.length >= 2) {
+        if (data_.noGrouping) {
+          if (data_.hasNA) {
+            endNumIndex = opts_.xDomain.length - 2 ;
+          }
         } else {
-          startNumIndex = 1;
-          if (data_.hasNA) {// has 'NA' tick
-            endNumIndex = opts_.xDomain.length - 3;
+          if (logScaleChecked) {
+            if (data_.hasNA) { // has 'NA' tick
+              endNumIndex = opts_.xDomain.length - 2;
+            }
           } else {
-            endNumIndex = opts_.xDomain.length - 2;
+            startNumIndex = 1;
+            if (data_.hasNA) {// has 'NA' tick
+              endNumIndex = opts_.xDomain.length >= 3 ? opts_.xDomain.length - 3 : endNumIndex;
+            } else {
+              endNumIndex = opts_.xDomain.length - 2;
+            }
           }
         }
       }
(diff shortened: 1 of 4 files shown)


app/scripts/views/components/barChart/barChart.js
```diff
@@ -21,6 +21,7 @@
         var val = d[data_.attrId];
         var _min;
         var _max;
+        var endNumIndex = opts_.xDomain.length - 1; // when data doesn't have NA
```
zhx828, line 24: What happened if xDomain is empty array?

app/scripts/views/components/barChart/barChart.js
```diff
@@ -44,12 +45,15 @@
...
+              if (data_.hasNA) {
+                endNumIndex = opts_.xDomain.length - 2;
+              } 
+              for (i = 2; i < endNumIndex; i++) {
                 if (d[data_.attrId] <= opts_.xDomain[i] &&
                   d[data_.attrId] >= opts_.xDomain[i - 1]) {
                   _min = opts_.xDomain[i - 1];
                   _max = opts_.xDomain[i];
-                  val = _min + (_max - _min) / 3;
+                  val = _min + (_max - _min) / 4;
```
zhx828, line 56: This should only be the case of linear scale?

app/scripts/views/components/barChart/barChart.js
```diff
@@ -295,87 +299,123 @@
...
-      var endNumIndex = 0;
+      var endNumIndex = opts_.xDomain.length - 1;
 
       tempFilters_[0] = '';
       tempFilters_[1] = '';
+      
+      // set start and end indexes for number bars
+      if (data_.noGrouping) {
+        if (data_.hasNA) {
+          endNumIndex = opts_.xDomain.length - 2;
```
zhx828: I'm curious since you didn't run into any issue on this, but does it matter to check xDomain length?

app/styles/partials/_global.scss
```diff
@@ -55,6 +55,12 @@
   top: 30%;
 }
 
+.error-panel {
+  position: absolute;
+  padding-left: calc(50% - 50px);
```
zhx828: This only works for text has 100px length

zhx828 (changes requested):

app/scripts/views/components/barChart/barChart.js
victoria34: If xDomain is empty array, it means no ticks show on the chart. No matter how users select, it always shows "Invalid Selection".

app/scripts/views/components/barChart/barChart.js
victoria34: Yes, we should check the xDomain length.

app/scripts/views/components/barChart/barChart.js
victoria34: If xDomain is empty array, it means no ticks show on the chart. No matter how users select, it always shows "Invalid Selection".

app/scripts/views/components/barChart/barChart.js
victoria34: Yes, I should check the xDomain length.

app/scripts/views/components/barChart/barChart.js
victoria34: I change it because I make gutter smaller when data is small number, which leads to bars become continuous. So It should be divided by 4 to separate bars.

app/scripts/views/components/barChart/barChart.js
```diff
@@ -306,27 +306,29 @@
       var hasGreaterOutlier = false;
       var hasSmallerOutlier = false;
       var startNumIndex = 0;
-      var endNumIndex = opts_.xDomain.length - 1;
+      var endNumIndex = opts_.xDomain.length >= 1 ? opts_.xDomain.length - 1 : 0;
```
zhx828: So if `opts_.xDomain` is empty, then endNumIndex is 0, then what's gonna happen at line 359?

app/styles/partials/_global.scss
```diff
@@ -56,8 +56,8 @@
 }
 
 .error-panel {
+  width: 100%;
   position: absolute;
-  padding-left: calc(50% - 50px); 
   padding-top: calc(50% - 50px);
```
zhx828: This doesn't work if the error message has two lines

app/scripts/views/components/barChart/barChart.js
victoria34: It won't run into line 359. Opts_.xDomain is empty means that opts_.xBarValues is also empty, so it won't run line 351 ~ 365.

zhx828 (approved):

PR #621, opened 2017-09-27
Copy share-virtual-cohort component to iViz.

victoria34:


app/index.html
```diff
@@ -124,6 +125,8 @@
         </div>
         <save-virtual-study :show-save-button="showSaveButton" :stats="stats"
```
zhx828, line 126: save-virtual-study should be disabled now

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+        _self.updateStats = true;
+        _self.$nextTick(function() {
+          _self.addNewVC = true;
+        });
+      }
+    },
+    ready: function() {
+      var self_ = this;
+      var previousSelectedCases = {};
+      if (this.showShareButton) {
```
zhx828: Should we give a default value to `showShareButton`? Probably we should do the same in showSaveButton

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+        $('.share-virtual-study').qtip({
```
zhx828: Do you think we can wrap up the qtip configuration to util? I think most of qtips are sharing the same configuration, only the content is different.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+              var tooltip = $('.iviz-share-cohort-btn-qtip .qtip-content');
+
+              tooltip.find('.share-cohort').click(function() {
+                tooltip.find('.shared').css('display', 'none');
+                tooltip.find('.dialog').css('display', 'none');
+                tooltip.find('.saving').css('display', 'none');
+                api.reposition();
+
+                // Copy virtual study link to clipboard
+                var $temp = $("<input>");
```
zhx828: Let's not using dollar sign in variant naming. It is very confusing with other variables from jQuery and vue native function.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+                $("body").append($temp);
+                $temp.val(tooltip.find('.virtual-study-link').attr('href')).select();
+                document.execCommand("copy");
```
zhx828: Could you give me a quick background on this line?

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+                self_.$nextTick(function() {
+                  var saveCohort = false;
+
+                  if (_.isObject(self_.stats.selectedCases)) {
+                    var selectedCasesMap = {};
+                    _.each(self_.stats.selectedCases, function(study){
+                      selectedCasesMap[study.studyID] = study;
+                    });
+
+                    // When users double click "Share" button, the virtual cohort will
```
zhx828: Why do we even allow user double click?

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,206 @@
...
+
+              var cohortName = $('#study_name').val();
+              var cohortDescription = $('#study_desc').val();
+              if (_.isObject(vcSession)) {
+                self_.updateStats = true;
+
+                self_.$nextTick(function() {
+                  var saveCohort = false;
+
+                  if (_.isObject(self_.stats.selectedCases)) {
```
zhx828, line 115: There should be validation on self_.stats. Probably in the vue properties, we need to specify this is an object.

app/scripts/vueCore.js
```diff
@@ -30,6 +30,7 @@
             charts: {},
             groupCount: 0,
             updateSpecialCharts: false,
+            showShareButton: true,
             showSaveButton: true,
```
zhx828: Disable this.

zhx828 (changes requested):

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: Sorry, I didn't explain clear it clearly. Specifically, it's not double click. When a user clicks copy, it will trigger saving the current virtual cohort and return the url to the user. When a user want to see the cohort url, he/she needs to click `Share` button. We always show the url to user but we don't need to same virtual cohort every time if it is same with the previous saved cohort.

app/scripts/vueCore.js
victoria34: I thought the default value of showShareButton is true and we can manipulate it index.html by setting iViz.vue.manage.getInstance().showShareButton = true/false. Right?

app/index.html
victoria34: I disabled it by setting iViz.vue.manage.getInstance().showSaveButton = false and it won't be displayed on page.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: I think the default value of showShareButton is true, which is set in vueCore.js right?

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: Sure, I will change it.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: Sorry, I'll leave comment for it.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: I'm not sure since html code of saveButton and shareButton are different.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
victoria34: Good point. Thanks!

app/index.html
zhx828: I'm wondering since we are not using the save button, should we comment out the whole function. But I guess configuration should work too.

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
zhx828: The qTio configuration code is almost identical, we can probably abstract that part.

app/scripts/vueCore.js
zhx828: Let's change the default of saving to false.

app/index.html
```diff
@@ -202,8 +205,13 @@
...
 
-    // This is used to indicate how to disable two buttons. By default, they are set to false.
-    iViz.vue.manage.getInstance().showSaveButton = true;
+    // This is used to indicate how to disable three buttons. By default, they are set to false.
+    iViz.vue.manage.getInstance().showSaveButton = false;
+
+    //Set this to true if want to share cohorts url(this is for cbioportal)
+    iViz.vue.manage.getInstance().showShareButton = true;
+
+    iViz.vue.manage.getInstance().showManageButton = true;
```
zhx828: Do we need the manage button?

app/scripts/views/components/shareVirtualStudy/shareVirtualStudy.js
```diff
@@ -0,0 +1,225 @@
...
+      return {
+        savedVC: null
+      };
+    },
+    watch: {
+      'showShareButton': function(showShareButton) {
+        // In case static qtip will be created multiple times.
+        if (showShareButton && !this.createdQtip) {
+          this.createQtip();
+          this.createdQtip = true;
```
zhx828: this should reside in createQtip method

zhx828 (approved):

PR #625, opened 2017-10-02
Create unit test for getTickFormat()

### LINKED ISSUE #623: Bar chart ticks reversed
http://localhost:8080/cbioportal/study?id=sarc_mskcc#summary

image

victoria34:
Created unit test for getTickFormat() in util.js

image

Fix #623


test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+var logScale = false;
+var data = {};
+var opts = {};
+var ticks = [];
+
+// noGrouping is true when number of ticks < 6.
+// logScale and noGrouping cannot be true at same time.
+// emptyValue only has value when hasNA = true.
+
+describe("cbio-util.js getTickFormat() tests", function () {
```
zhx828: cbio-util.js?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    data.min = _.min(opts.xDomain);
+    data.max = _.max(opts.xDomain);
+
+    _.each(opts.xDomain, function(v) {
+      ticks.push(util.getTickFormat(v, logScale, data, opts));
+    });
+
+    // console.log(opts.xDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length);
```
zhx828: For the length check, this should be done separately. Test some extreme cases.

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    };
+
+    data.min = _.min(opts.xDomain);
+    data.max = _.max(opts.xDomain);
+
+    _.each(opts.xDomain, function(v) {
+      ticks.push(util.getTickFormat(v, logScale, data, opts));
+    });
+
+    // console.log(opts.xDomain, ticks);
```
zhx828: Could you remove all commented out code?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+      } else {
+        expect(ticks[i]).to.be.equal(ticks[i]);
+      }
+    }
+  });
+
+  it('3. Has NA and outliers.', function() {
+    ticks = [];
+    logScale = false;
+    data = {
```
zhx828: If the testing is for the `data`. Shouldn't we use the same xDomian. Make the data be only variable in different tests?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    });
+
+    // console.log(opts.xDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length); // same length
+    expect(ticks[0]).to.be.equal('<=' + opts.xDomain[1]); // first tick = '<=' + value of second tick
+    expect(ticks[ticks.length - 1]).to.be.equal('>' + opts.xDomain[ticks.length - 2]); // last tick = '>' + value of last second tick
+  });
+
+  it('2. logScale but no NA tick. Odd tick should start from 1***, even tick should be "". ', function() {
```
zhx828: I think in this test, you only checked if the even ticks are empty string, but you didn't check the odd values are fractional exponent power of 10

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+      'noGrouping': true,
+      'hasNA': false,
+      'min': '',
+      'max': ''
+    };
+
+    opts = {
+      'xDomain' : [3, 5, 6, 9],
+      'emptyMappingVal' : '',
+      'xFakeDomain': [0, 1, 2, 3]
```
zhx828: In there, you introduce a new concept, `xFakeDomain`. But this test is designed for data.noGrouping? Do we have to add `xFakeDomain` in order to test noGrouping?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    // console.log(opts.xDomain, opts.xFakeDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length);
+    for (var i = 0; i < opts.xDomain.length - 1; i++) {
+      expect(ticks[i]).to.be.equal(opts.xDomain[i]);
+    }
+    expect(ticks[ticks.length - 1]).to.be.equal('NA'); // last tick = 'NA' 
+  });
+
+  it('5. Ticks should be same as xDomain (no max and min outlier) when no grouping is true except last tick(NA).', function() {
```
zhx828: This description is the same with 4. You need to describe by last tick is NA here

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    // console.log(opts.xDomain, opts.xFakeDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length);
+    for (var i = 0; i < opts.xDomain.length - 1; i++) {
+      expect(ticks[i]).to.be.equal(opts.xDomain[i]);
+    }
+    expect(ticks[ticks.length - 1]).to.be.equal('NA'); // last tick = 'NA' 
+  });
+
+  it('6. logScale and has NA tick. Odd tick should start from 1***, even tick should be "" and last tick is "NA".', function() {
```
zhx828: I think in overall, all descriptions need to be clear on what needs to be tested under which condition. If this is not the case, please let me know.

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    logScale = true;
+    data = {
+      'noGrouping': false,
+      'hasNA': false,
+      'min': '',
+      'max': ''
+    };
+
+    opts = {
+      'xDomain' : [1, 3, 10, 31, 100, 316, 1000, 3162, 10000, 31622],
```
zhx828: I think it would be great to put all preconditions into a variable pool in order to avoid same code

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+      if ((i + 1) % 2 === 0) {
+        expect(ticks[i]).to.be.equal('');
+      } else {
+        expect(ticks[i]).to.be.equal(ticks[i]);
+      }
+    }
+    expect(ticks[ticks.length - 1]).to.be.equal('NA'); // last tick = 'NA' 
+  });
+
+  it('1.3 More than 5 ticks when no NA and has outliers.', function() {
```
zhx828: why this test number becomes 1.3 after 6?

zhx828: How do you define `has outliers`?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    });
+
+    // console.log(opts.xDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length); // same length
+    expect(ticks[0]).to.be.equal('<=' + opts.xDomain[1]); // first tick = '<=' + value of second tick
+    expect(ticks[ticks.length - 1]).to.be.equal('>' + opts.xDomain[opts.xDomain.length - 2]); // last tick = '>' + value of last second tick
+  });
+
+  it('7. Special case for year: min year > 1500, and length > 7.', function() {
```
zhx828: I can not understand the test from the description. Why 1500 and 7?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    });
+
+    // console.log(opts.xDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length); // same length
+    expect(ticks[0]).to.be.equal('<=' + opts.xDomain[1]); // first tick = '<=' + value of second tick
+    expect(ticks[ticks.length - 1]).to.be.equal('>' + opts.xDomain[opts.xDomain.length - 2]); // last tick = '>' + value of last second tick
+  });
+
+  it('7.2 Special case for year; min year < 1500, and length > 7.', function() {
```
zhx828: I think we mentioned once on setting up the threshold on the year. Could you remind me why we need to do that?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    data.min = _.min(opts.xDomain);
+    data.max = _.max(opts.xDomain);
+
+    _.each(opts.xDomain, function(v) {
+      ticks.push(util.getTickFormat(v, logScale, data, opts));
+    });
+
+    // console.log(opts.xDomain, ticks);
+
+    expect(ticks).to.have.lengthOf(opts.xDomain.length); // same length
```
zhx828: You don't want to do any tests except than the length?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
```
zhx828: No test on here?

test/unit/getTickFormat.spec.js
```diff
@@ -0,0 +1,667 @@
...
+    expect(ticks[ticks.length - 1]).to.be.equal('NA'); // last tick = 'NA' 
+  });
+
+  it('8.2.2 Small Data: initial of ticks is not "1", has NA and noGrouping = false\'', function() {
```
zhx828: What's the difference between 8.2.1 and 8.2.2 except the xDomains are different? If you are trying to test the algorithm can accept the different input, could you give some more information how the input is designed? Why the test is needed?

zhx828 (changes requested):

test/unit/getTickFormat.spec.js
victoria34: I didn't created the threshold but I guess it's used for label of ticks not overlapping.

zhx828 (approved):
@victoria34 LGTM, could you solve the conflicts?

PR #635, opened 2017-11-15
Switch to new api & performance improvment

zhx828:
1. Switch to new POST APIs
2. Performance improvement
    - Remove iViz.util.strIsNa check, isNaN can handle most of the cases
    - Calculate dataMetaKeys in object instead of using _.unique
    - Replace jQuery each with underscore each
    - Only calculate hasMutationData/hasSnaSegmentData once
    - Remember which study contains clinical attribute
    - Sort clinical attribute based on # of studies first
    - Calculate whether clinical attributes is preselected attr by regex,
this is O(n) instead of previous O(n2)
    - Do not sort clinical attrs when first time initializes result.groups.
They will be calculated in the following step
    - Update preSelectClinicalAttrRegex, make it more efficient
    - Calculate group attrs only once
    - No need to call strIsNa since data all have been normalization to NA
    - Optimize survival plot when # of dots more than 5000, decrease
the number of patients used generating survival plot. The threshold is
sqrt(pow(time max- time min) + pow(survival rate max- survival rate
min))

### DIFF
--- a/app/scripts/model/dataProxy.js
+++ b/app/scripts/model/dataProxy.js
@@ -827,6 +827,7 @@ window.DataManagerForIviz = (function($, _) {
                 _profile.push({
                   id: profile.molecularProfileId,
                   study_id: profile.studyId,
+                  datatype: profile.datatype,
                   genetic_alteration_type: profile.molecularAlterationType
                 });
                 profiles[profile.studyId] = _profile;
@@ -898,10 +899,10 @@ window.DataManagerForIviz = (function($, _) {
                   description: attribute.description,
                   display_name: attribute.displayName,
                   is_patient_attribute: attribute.patientAttribute ? "1" : "0",
-                  prority: attribute.priority
+                  priority: attribute.priority
                 });
                 attributes[attribute.studyId] = _attribute;
-              });c
+              });
               var selectedAttributes = _.pick(attributes, self.getCancerStudyIds());
               var clinical_attributes_set = {};
               _.each(selectedAttributes, function(studyAttributes, studyId) {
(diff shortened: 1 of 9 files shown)


zhx828:
@jjgao Could you also help me verify the logic on the survival optimization? Thanks.
With all patients:
full

With partial patients under 200
```
var timeThreshold = (_dataTimeMax - _dataTimeMin) / 200;
var survivalRateThreshold = (_dataSurvivalRateMax - _dataSurvivalRateeMin) / 200;
var averageThreshold = iViz.util.getHypotenuse(timeThreshold, survivalRateThreshold);
```
200-threshold

zhx828:
This pull request depends on https://github.com/cBioPortal/cbioportal/pull/3366

app/scripts/model/dataProxy.js
```diff
@@ -97,22 +97,23 @@ window.DataManagerForIviz = (function($, _) {
               var _cnaCaseUIDs = [];
               var _sequencedCaseUIDs = [];
               var _allCaseUIDs = [];
+              var _allStudyIds = self.getCancerStudyIds();
 
               iViz.priorityManager.setDefaultClinicalAttrPriorities(_configs.priority);
 
-              $.each(_caseLists, function(studyId, caseList) {
+              _.each(_caseLists, function(caseList, studyId) {
                 if (caseList.cnaSampleIds.length > 0) {
```
kalletlak: checking length is not required here. If possible could you clean this.

zhx828:
Sure thing.

kalletlak:
Lot of cleanup can be done in `dataproxy.js`. we handle this in another PR. Apart from that everything looks good to me

kalletlak (approved):

PR #639, opened 2017-11-17
Update virtual study name and description

zhx828:


app/scripts/session/sessionUtil.js
```diff
@@ -50,17 +50,10 @@
...
-        def.resolve(_virtualCohort);
-      } else {
-        $.when(generateCohortDescription_(cases)).done(function(_desp) {
-          _virtualCohort.description = _desp;
-          def.resolve(_virtualCohort);
-        });
+        _virtualCohort.name = "Selected " + (cases.length > 1 ? "Studies" : "Study");
       }
+      _virtualCohort.description = description || '';
+      def.resolve(_virtualCohort);
```
kalletlak, line 56: Deferring would no longer be required here. May be we need to clean this in next release

kalletlak (approved):

app/scripts/session/sessionUtil.js
zhx828: It would tricky/unworthy to do to clean up all the code since the return type changed. Will clean up in next release.

PR #640, opened 2017-11-17
Performance improvement

zhx828:
Only sort caseUids once;
Use sorted difference method;
Use nonNaCases.push.apply instead of concat. concat creates new array
which uses lots of memory and causing GC. Based on
https://stackoverflow.com/questions/5080028/what-is-the-most-efficient-w
ay-to-concatenate-n-arrays-in-javascript.

### DIFF
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -342,7 +342,7 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
             _mutGeneMeta[_uniqueId] = {};
             _mutGeneMeta[_uniqueId].gene = _uniqueId;
             _mutGeneMeta[_uniqueId].num_muts = 1;
-            _mutGeneMeta[_uniqueId].case_uids = [_caseUIdIndex];
+            _mutGeneMeta[_uniqueId].case_ids = [_caseUIdIndex];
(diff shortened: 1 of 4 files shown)


app/scripts/main.js
```diff
@@ -342,7 +342,7 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
             _mutGeneMeta[_uniqueId] = {};
             _mutGeneMeta[_uniqueId].gene = _uniqueId;
             _mutGeneMeta[_uniqueId].num_muts = 1;
-            _mutGeneMeta[_uniqueId].case_uids = [_caseUIdIndex];
+            _mutGeneMeta[_uniqueId].case_ids = [_caseUIdIndex];
```
kalletlak, line 345: any specific reason for rename ?

zhx828: Keep two sets. uids will only need to be calculated once

app/scripts/views/components/dataTable/tableView.js
```diff
@@ -298,7 +303,7 @@
           var freq = 0;
           datum.gene = item.gene;
           if (_selectedGenesMap === undefined) {
-            datum.case_uids = iViz.util.unique(item.case_uids);
+            datum.case_uids = item.case_uids;
```
kalletlak, line 306: Are you sure `item.case_uids` be unique?

kalletlak (approved):

app/scripts/views/components/dataTable/tableView.js
zhx828: yes, uids will be unique cases

PR #650, opened 2017-12-08
Uppercase all clinical attributes from server

zhx828:
All upper case has been used as convention at the server side, but sometimes it may not be strictly followed. The changes in this pull request prevent the issue happening.


kalletlak (approved):

kalletlak:
@zhx828 make sure to add a comment at the changes. Would be useful during refactoring once its fixed at the backend.

PR #652, opened 2017-12-19
Changes wrt saving virtual study

kalletlak:
These changes enables to save user virtual studies instead of just sharing them. This would be useful for secured portal which have some authentication. This enables logged in user to see their virtual studies in home page
This is related to https://github.com/cBioPortal/cbioportal/issues/3450

### DIFF
--- a/app/scripts/session/sessionEvent.js
+++ b/app/scripts/session/sessionEvent.js
@@ -9,7 +9,7 @@ window.vcSession = window.vcSession ? window.vcSession : {};
   }
   vcSession.events = (function() {
     return {
-      saveCohort: function(stats, name, description, addToUserStudies = false) {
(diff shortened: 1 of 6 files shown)


jjgao:
@kalletlak could you please add some description?

app/scripts/session/sessionEvent.js
```diff
@@ -9,11 +9,11 @@ window.vcSession = window.vcSession ? window.vcSession : {};
   }
   vcSession.events = (function() {
     return {
-      saveCohort: function(stats, name, description) {
+      saveCohort: function(stats, name, description, addToUserStudies = false) {
```
zhx828: This is ES6 format, we don't support that yet.

zhx828 (approved):

PR #656, opened 2018-01-04
Support study_id:sample_id in Custom case selection box

zhx828:
- This pull request intends to solve issue https://github.com/cBioPortal/cbioportal/issues/3620
- Pre-fill the box with the current selected sample/patient

### DIFF
--- a/app/scripts/vueCore.js
+++ b/app/scripts/vueCore.js
@@ -269,26 +269,25 @@
               var unmappedCaseIDs = [];
 
               _.each(selectedCases, function(id) {
-                if (id) {
-                  var caseUIDs = [];
-                  if (id.indexOf(':') !== -1) {
-                    var pair = id.split(':').map(function(t) {
-                      return t.trim();
-                    });
-                    if (pair.length == 2) {
-                      var caseId = iViz.getCaseIndex(selectionType, pair[0], pair[1]);
-                      if (caseId) {
-                        caseUIDs.push(caseId);
-                      }
-                    }
-                  } else {
-                    caseUIDs = iViz.getCaseUID(selectionType, id);
+                var caseUIDs = [];
+                var pair = id
+                  .split(':')
+                  .map(function(t) {
+                    return t.trim();
+                  });
+                if (pair.length == 2) {
+                  var caseId = iViz.getCaseIndex(selectionType, pair[0], pair[1]);
+                  if (caseId) {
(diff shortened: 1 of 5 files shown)


app/scripts/vueCore.js
```diff
@@ -269,11 +269,26 @@
               var unmappedCaseIDs = [];
 
               _.each(selectedCases, function(id) {
-                var caseUIDs = iViz.getCaseUID(selectionType, id);
-                if (caseUIDs.length === 0) {
-                  unmappedCaseIDs.push(id);
-                } else {
-                  selectedCaseUIDs = selectedCaseUIDs.concat(caseUIDs);
+                if (id) {
```
kalletlak: why is this `if` condition required here

app/scripts/vueCore.js
```diff
@@ -269,11 +269,26 @@
...
+                  var caseUIDs = [];
+                  if (id.indexOf(':') !== -1) {
```
kalletlak: instead of having this condition why don't we just split and check the length. something like
```
var pair = id
            .split(':')
            .map(function(t) {
                return t.trim();
            });
if (pair.length == 2) {
  var caseId = iViz.getCaseIndex(selectionType, pair[0], pair[1]);
  if (caseId) {
    caseUIDs.push(caseId);
  }
}
else {
  caseUIDs = iViz.getCaseUID(selectionType, id);
}
```

app/scripts/vueCore.js
```diff
@@ -269,11 +269,26 @@
...
+                      return t.trim();
+                    });
+                    if (pair.length == 2) {
+                      var caseId = iViz.getCaseIndex(selectionType, pair[0], pair[1]);
+                      if (caseId) {
+                        caseUIDs.push(caseId);
+                      }
+                    }
+                  } else {
+                    caseUIDs = iViz.getCaseUID(selectionType, id);
```
kalletlak: I think we could change `iViz.getCaseUID` to return single id instead of array now

zhx828: Well, it is still a valid function though and if you have the same id in multiple studies, it should return an array.

app/scripts/vueCore.js
zhx828: Hm... after second thought, it shouldn't be there. But I remember I did encounter to the undefined issue. Remove for now.

kalletlak (approved):

PR #657, opened 2018-01-05
Don't post case ids if there is no filter applied

zhx828:


app/_components/QueryByGeneUtil.js
```diff
@@ -19,25 +19,35 @@ window.QueryByGeneUtil = (function() {
...
       var _arr = [];
-      _.each(selectedCases, function(_obj) {
-        var _studyId = _obj.id;
-        _.each(_obj.samples, function(_sampleId) {
-          _arr.push(_studyId + ":" + _sampleId);
+      var formOps = {
+        cancer_study_list: studyIds
+      };
+
+      if (_.isObject(selectedCases)) {
```
kalletlak: could you make sure `selectedCases` is null when no filters applied.

app/scripts/main.js
```diff
@@ -658,13 +658,13 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
 
       if (_self.cohorts_.length === 1) { // to query single study
         if (QueryByGeneTextArea.isEmpty()) {
-          QueryByGeneUtil.toMainPage(_self.cohorts_[0], _self.stat().studies);
+          QueryByGeneUtil.toMainPage(_self.cohorts_[0], vm_.hasfilters ? _self.stat().studies : undefined);
```
kalletlak: Question, should `cancer_study_list` be input studies or selected studies after applying filters?

app/_components/QueryByGeneUtil.js
```diff
@@ -19,25 +19,35 @@ window.QueryByGeneUtil = (function() {
...
-      });
-      submitForm(window.cbioURL + 'index.do', {
-        cancer_study_list: selectedCases.map(function(x) {return x.id}),
-        cancer_study_id: 'all',
-        case_set_id: -1,
-        case_ids: _arr.join('+')
-      });
+
+        formOps = {
+          cancer_study_list: selectedCases.map(function(t) {
```
kalletlak, line 42: I think `selectedCases.map(study => study.id) would be cleaner

zhx828: Are we doing es6 going forward?

app/scripts/main.js
zhx828: selected studies after applying filters

kalletlak: Sorry, I got this. If no filters applied first parameter would be used else studies from stat would be used.

kalletlak: instead of passing `undefined` and checking it, why not just pass `_self.stat().studies` and `hasfilters` flag.

kalletlak (approved):

PR #660, opened 2018-01-11
Fix error while saving virtual study

kalletlak:


app/scripts/session/sessionUtil.js
```diff
@@ -46,7 +46,13 @@
       var def = new $.Deferred();
       var _virtualCohort = $.extend(true, {}, virtualCohort_);
       _virtualCohort.filters = filters;
-      _virtualCohort.studies = cases;
+      
+      _virtualCohort.studies = cases.map(function(studyObj) {
+        var _study = $.extend(true, {}, studies_);
```
zhx828: If you manually assign id and samples to _study object. No need to do a deep copy anymore. `studies_` can even be removed

kalletlak: yeah! true

zhx828 (approved):

PR #665, opened 2018-01-25
Fix the view linkout 

zhx828:


kalletlak (approved):
This is ok for now as this page would be refactored to react in future

PR #677, opened 2018-01-31
Include typescript and grunt-ts

zhx828:


zhx828:
We decided to develop React components in portal front-end repo. The typescript integration is no longer needed here.

PR #681, opened 2018-02-08
 Merge Save/Share virtual study button together into Virtual Study button

zhx828:


app/index.html
```diff
@@ -125,10 +124,7 @@
             <span>{{ selectedSamplesNum }} samples / {{ selectedPatientsNum }} patients</span>
           </span>
         </div>
-        <save-virtual-study :show-save-button="showSaveButton" :stats="stats"
-                           :update-stats.sync="updateStats"></save-virtual-study>
-        <share-virtual-study :show-share-button="showShareButton" :stats="stats"
-                             :update-stats.sync="updateStats"></share-virtual-study>
+        <virtual-study v-if="showVirtualStudyButton" :stats="stats" :update-stats.sync="updateStats"></virtual-study>
```
kalletlak: Doesn't we require two control parameters each for share and save button?. Or we can show share button for default and a controlled save button.

app/scripts/model/dataProxy.js
```diff
@@ -799,6 +799,7 @@ window.DataManagerForIviz = (function($, _) {
         return window.cbio.util.deepCopyObject(this.studyCasesMap);
       },
       data: {
+        studies: {},
```
kalletlak, line 802: Are these changes for this PR. were is it used?

kalletlak (changes requested):

app/scripts/model/dataProxy.js
zhx828: Actually I included a fix in this pull request, it caches the study meta info instead of fetching the data everytime when we try to get display name see line 1599-1631 in dataproxy

app/index.html
zhx828: After discussion, I will bring it back.

app/scripts/views/components/virtualStudy/virtualStudy.js
```diff
@@ -0,0 +1,353 @@
...
+        savedVS: null,
+        virtualStudyName: 'Selected Study',
+        virtualStudyDescription: ''
+      };
+    },
+    methods: {
+      styleDisplay: function(el, action) {
+        el.css('display', action);
+      },
+      showDialog: function(tooltip) {
```
kalletlak, line 36: I would prefer to go with one function and pass more parameters `className` and `display`

app/scripts/views/components/virtualStudy/virtualStudy.js
```diff
@@ -0,0 +1,353 @@
...
+                });
+              this.createdQtip = true;
+            },
+            show: function() {
+              var tooltip = $('.iviz-virtual-study-btn-qtip .qtip-content');
+              self_.updateStats = true;
+              self_.$nextTick(function() {
+                // If user hasn't specific name only.
+                if (tooltip.find('.cohort-name').val() === '') {
+                  tooltip.find('.cohort-name').val(self_.virtualStudyName);
```
kalletlak, line 281: May be just a placeholder when nothing is set?

app/scripts/views/components/virtualStudy/virtualStudy.js
zhx828: I actually prefer the current implementation. The logic is clear. We don't need to remember where are places using this function and change the class name individually if that happens

app/scripts/views/components/virtualStudy/virtualStudy.js
zhx828: I thought we agreed to use the default name as always?

kalletlak: true, but if user want's to change the name, he needs to delete the existing name and enter new one. instead check it at when the save button is clicked.

zhx828: I'm just afraid the placeholder is not obvious enough letting them know this will be saved if no named specified. Again, we shouldn't spend too much time discussing the details about this. Everything will change after sending out the testing website.

kalletlak (approved):

PR #682, opened 2018-02-08
EFDT update; Give message after adding chart; 

zhx828:
Add tooltip for all icons in chart operations header
This closes https://github.com/cBioPortal/cbioportal/issues/3853
Add notification after adding a chart
Assign table component with an element Id

### DIFF
--- a/app/_components/MutatedGeneCNATable.js
+++ b/app/_components/MutatedGeneCNATable.js
@@ -1480,7 +1480,7 @@ window.EnhancedFixedDataTableSpecial = (function() {
         selectedGene: [],
         sortBy: 'cases',
         sortDir: 'DESC',
-        elementId: '',
+        elementId: 'EFDT',
         isResizable: false
       };
     },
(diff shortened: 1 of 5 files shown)


app/_components/MutatedGeneCNATable.js
```diff
@@ -578,16 +579,17 @@ window.EnhancedFixedDataTableSpecial = (function() {
 
     // Creates Qtip after page scrolling
     onScrollEnd: function() {
-      $(".qtip").remove();
+      var qtipId = '.qtip';
+      if (this.props.elementId) {
+        qtipId = '.' + this.props.elementId + '-qtip ' + qtipId;
+      }
+      $(qtipId).remove();
```
kalletlak, line 586: what if `this.props.elementId` is `undefined`? it runs `$(".qtip").remove();` would that be ok?

kalletlak (approved):

app/_components/MutatedGeneCNATable.js
zhx828: Currently I set the property default elementId to empty String. What about I set it to `EFDT`? It would be better only remove qtip in EFDT tables?

PR #687, opened 2018-02-19
Cleanup

kalletlak:
1. remove vs-session.js
2. add origin to virtual study while saving/sharing

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -70,13 +70,10 @@
   <script src="_components/cbioportal-client.js"></script>
   <!-- endbuild -->
 
-  <!-- build:js scripts/vs-session.js -->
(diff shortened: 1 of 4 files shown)


app/index.html
```diff
@@ -70,13 +70,10 @@
   <script src="_components/cbioportal-client.js"></script>
   <!-- endbuild -->
 
-  <!-- build:js scripts/vs-session.js -->
+  <!-- build:js scripts/iviz.js -->
   <script src="scripts/session/sessionEvent.js"></script>
```
zhx828, line 74: Are we no longer using session on other pages?

kalletlak: for the frontend repo we are having its own typescript client

PR #689, opened 2018-02-21
Fix issues when querying from study view

kalletlak:


app/index.html
```diff
@@ -222,15 +219,58 @@
     window.cohortIdsList = ['hnsc_tcga'];
     window.iviz = {};
     window.iviz.datamanager = new DataManagerForIviz.init(window.cbioURL, studyCasesMap);
+    var getSelectableStudyIds = function() {
+      var def = new $.Deferred();
+      $.when(this.getPhysicalStudies(), this.getVirtualStudies()).then(function(physicalStudies, virtualStudies) {
+        var physicalStudyIds = _.pluck(physicalStudies,'studyId');
```
zhx828: `physicalStudyIds ` is not used anywhere?

app/index.html
```diff
@@ -222,15 +219,58 @@
...
+          }
+        });
+        virtualStudyIds = _.flatten(virtualStudyIds);
+        def.resolve(virtualStudyIds)
+      }).fail(function(error) {
+        def.reject(error);
+      });
+      return def.promise();
+    }
+    var getVirtualStudies = function(){
```
zhx828: Just curious for future development, shouldn't this function access user email? Or the proxy endpoint will actually fetch studies for this user only?

app/index.html
```diff
@@ -222,15 +219,58 @@
...
         if (_.isObject(configs)) {
           opts = configs;
         }
         $.when(
-          window.iviz.datamanager.initialSetup()
-        ).done(function(_data) {
+          window.iviz.datamanager.initialSetup(),
+          getSelectableStudyIds()
+        ).done(function(_data,selectableIds) {
           iViz.init(_data, opts);
```
zhx828: `selectableIds` hasn't been passed into `init` yet

app/scripts/main.js
```diff
@@ -45,8 +46,20 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
       vm_ = iViz.vue.manage.getInstance();
+      var selectableIdsSet = {}
+      _.each(_selectableIds, function(id){
+        selectableIdsSet[id] = true;
+      });
+
+      var cohortIds = window.cohortIdsList;
+      for (var i = 0; i < cohortIds.length; i++) {
+        if(selectableIdsSet[cohortIds[i]]){
+          includeCases = false;
```
zhx828, line 59: This will only be true is cohortIdsList is empty?

app/_components/QueryByGeneUtil.js
```diff
@@ -24,76 +24,54 @@ window.QueryByGeneUtil = (function() {
...
+    * geneIds : selected genes
+    * includeCases : indicates whether to include custom samples in the form.
+    *                this is true for shared virtual study
+    */
+    query: function(cohortIds, stats, geneIds, includeCases) {
       var formOps = {
-        cancer_study_list: studyIds
-      };
+        cancer_study_list: cohortIds,
+        cancer_study_id:'all'
```
zhx828, line 38: In the original toMainPage function, cancer_study_id is not always `all`

app/_components/QueryByGeneUtil.js
```diff
@@ -24,76 +24,54 @@ window.QueryByGeneUtil = (function() {
...
-
-        formOps = {
-          cancer_study_list: selectedCases.map(function(t) {
-            return t.id
-          }),
-          cancer_study_id: 'all',
-          case_set_id: -1,
-          case_ids: _arr.join('+')
-        };
+        formOps['case_set_id'] = -1;
```
zhx828, line 68: In original implementation, `case_set_id` is always -1

app/_components/QueryByGeneUtil.js
```diff
@@ -24,76 +24,54 @@ window.QueryByGeneUtil = (function() {
...
-      submitForm(window.cbioURL + 'index.do', {
-        cancer_study_list: selectedCases.map(function(x) {
-          return x.id
-        }),
-        cancer_study_id: 'all',
-        case_ids: _arr.join('+'),
-        case_set_id: -1,
-        gene_set_choice: 'user-defined-list',
-        gene_list: encodeURIComponent(selectedGenes),
-        Z_SCORE_THRESHOLD: 2.0,
```
zhx828, line 70: This is gone?

app/_components/QueryByGeneUtil.js
```diff
@@ -24,76 +24,54 @@ window.QueryByGeneUtil = (function() {
...
-        genetic_profile_ids_PROFILE_MUTATION_EXTENDED: mutationProfileId,
-        genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION: cnaProfileId,
-        clinical_param_selection: null,
```
zhx828, line 73: This is gone?

app/_components/QueryByGeneUtil.js
```diff
@@ -24,76 +24,54 @@ window.QueryByGeneUtil = (function() {
...
-        });
-      });
-      submitForm(window.cbioURL + 'index.do', {
-        cancer_study_list: selectedCases.map(function(x) {
-          return x.id
-        }),
-        cancer_study_id: 'all',
-        case_ids: _arr.join('+'),
-        case_set_id: -1,
-        gene_set_choice: 'user-defined-list',
```
zhx828, line 68: This is gone?

zhx828 (changes requested):
Need a second round review after questions/issues solved

app/_components/QueryByGeneUtil.js
kalletlak: it should be ok setting it as `all` as the study ids would be present in `cancer_study_list`

app/_components/QueryByGeneUtil.js
kalletlak: `case_set_id` is -1 only for custom case set

app/_components/QueryByGeneUtil.js
kalletlak: I think by default it is 'user-defined-list'

app/_components/QueryByGeneUtil.js
kalletlak: even this is by default  `2.0`

app/_components/QueryByGeneUtil.js
kalletlak: I think this parameter is never used in back-end or front-end

app/index.html
kalletlak: proxy api takes care of the email

app/scripts/main.js
kalletlak: it will be true for shared VS

zhx828 (approved):

Issue #691, opened 2018-02-22
Name and Description in Study View


kalletlak:
| Query Type        | Name           | Description  |
| ------------- |-------------| -----|
| Single Physical Study      | Physical Study Name | Physical Study Description |
| Multiple Physical Studies      | Combined Study      |   Each Physical study names with link-outs to study view  |
| Shared Virtual Study |   |    |
| Saved Virtual Study |    |    |
| Combined Studies |   |    |

PR #695, opened 2018-02-22
Use default VS name and description for shared VS

zhx828:
This solves #691


app/scripts/session/sessionUtil.js
```diff
@@ -49,7 +49,10 @@
 
     return {
       buildVCObject: buildVCObject_,
-      generateCohortDescription: generateCohortDescription_
+      generateVSName: function() {
```
kalletlak: May be a constant instead of function

kalletlak (approved):

PR #697, opened 2018-02-28
Always send in selected samples when init react table

zhx828:
Gene panel map contains the full samples from the panel, but not all
samples are in current study, this solves
https://github.com/cBioPortal/cbioportal/issues/3908;
Do not after first time initialized the table, need to set initialized
to true;

### DIFF
--- a/app/scripts/views/components/dataTable/tableView.js
+++ b/app/scripts/views/components/dataTable/tableView.js
@@ -55,9 +55,9 @@
       function(_attributes, _opts, _selectedSamples, _selectedGenes,
                _data, _callbacks, _geneData, _dimension, _genePanelMap) {
         initialized = false;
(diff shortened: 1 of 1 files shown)


app/scripts/views/components/dataTable/tableView.js
```diff
@@ -76,7 +76,8 @@
           group = dimension.group();
           initPieTableData();
         }
-        initReactTable(true);
+        initReactTable(true, undefined, sequencedSampleIds);
```
kalletlak: instead of sending `undefined` as an parameter why not `initReactTable(true, sequencedSampleIds)` moving that `undefined` parameter to last in `initReactTable` implementation

kalletlak (approved):

PR #700, opened 2018-03-13
Fix <18 >89 age bin issue; Fix NA mutation counts after adding chart

zhx828:
Fixed the bugs
- https://github.com/cBioPortal/cbioportal/issues/3926
- https://github.com/cBioPortal/cbioportal/issues/3931

### DIFF
--- a/app/scripts/views/components/barChart/barChart.js
+++ b/app/scripts/views/components/barChart/barChart.js
@@ -16,8 +16,8 @@
       var tickVal = [];
       var i = 0;
       var barSet = {};
-      
-      dcDimension = ndx_.dimension(function (d) {
+
+      dcDimension = ndx_.dimension(function(d) {
         var val = d[data_.attrId];
         var _min;
         var _max;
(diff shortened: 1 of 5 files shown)


app/scripts/views/components/barChart/barChart.js
```diff
@@ -723,13 +731,13 @@
                     Number(cbio.util.toPrecision(Number(_tmpValue), 3, 0.1));
                 }
 
-                // If the current tmpValue already bigger than maxmium number, the
+                // If the current tmpValue already bigger than maximum number, the
                 // function should decrease the number of bars and also reset the
-                // Mappped empty value.
+                // Mapped empty value.
                 if (_tmpValue >= max) {
```
victoria34, line 737: Could we just change the if condition to `_tmpValue < max` and remove `break` ?

victoria34 (changes requested):

kalletlak (approved):

app/scripts/views/components/barChart/barChart.js
zhx828: We need to break the for loop if the current value is higher than the max

victoria34 (approved):

PR #708, opened 2018-03-28
Do not reverse genes, and avoid jQuery unique method

zhx828:
This solves issue https://github.com/cBioPortal/cbioportal/issues/3991

### DIFF
--- a/app/_components/QueryByGeneTextArea.js
+++ b/app/_components/QueryByGeneTextArea.js
@@ -12,9 +12,14 @@ var GenelistModel = Backbone.Model.extend({
         delim = delim || " ";
         return this.getCleanGeneArray().join(delim);
     },
-
-    getCleanGeneArray: function(){
-        return $.unique(this.removeEmptyElements(this.get("geneString").toUpperCase().split(/[^a-zA-Z0-9-]/))).reverse();
+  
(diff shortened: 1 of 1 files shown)


app/_components/QueryByGeneTextArea.js
```diff
@@ -12,9 +12,14 @@ var GenelistModel = Backbone.Model.extend({
...
         return this.getCleanGeneArray().join(delim);
     },
-
-    getCleanGeneArray: function(){
-        return $.unique(this.removeEmptyElements(this.get("geneString").toUpperCase().split(/[^a-zA-Z0-9-]/))).reverse();
+  
+    getCleanGeneArray: function() {
+      var _arr = this.removeEmptyElements(this.get("geneString")
+        .toUpperCase().split(/[^a-zA-Z0-9-]/));
+      _arr = _arr.filter(function(item, pos) {
```
kalletlak, line 19: any specific reason to filter?

zhx828: In the original implementation, there is a unique method. I don't want to touch more code in here since we are going to refactoring this component anyway.

kalletlak (approved):

PR #717, opened 2018-04-12
Use sample/patient UID's instead of creating one

kalletlak:


kalletlak:
Closing this are we would be migrating to new front-end prototype soon!

PR #718, opened 2018-04-12
Improve VS default decription

zhx828:
Include all filters in description;
This solves issue https://github.com/cBioPortal/cbioportal/issues/4080;
This solves issue https://github.com/cBioPortal/cbioportal/issues/4075;

### DIFF
--- a/app/scripts/main.js
+++ b/app/scripts/main.js
@@ -617,6 +617,16 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
     getSampleIds: function(studyId, patientId) {
       return data_.groups.group_mapping.studyMap[studyId].patient_to_sample[patientId];
     },
+    getStudyCacseIdsUsingUIDs: function(type, uids) {
+      var ids = [];
+      _.each(uids, function(uid) {
+        ids.push({
+          studyId: data_.groups[type].data[uid].study_id,
+          caseId: data_.groups[type].data[uid].sample_id
+        });
+      });
+      return ids;
+    },
     openCases: function() {
       var _selectedCasesMap = {};
       var _patientData = data_.groups.patient.data;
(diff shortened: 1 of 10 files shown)


app/scripts/session/sessionUtil.js
```diff
@@ -37,20 +37,59 @@
...
+            } else {
+              _desp += _filter.filter.join(', ');
+            }
+          });
+        }
+
+        _desp += '\n\nCreated on  ' + getCurrentDate();
+
+        if (window.userEmailAddress) {
+          _desp += ' by ' + window.userEmailAddress;
```
kalletlak, line 91: Why are we adding email id in description. we save email id for logged in users

app/scripts/main.js
```diff
@@ -776,13 +786,29 @@ window.iViz = (function(_, $, cbio, QueryByGeneUtil, QueryByGeneTextArea) {
...
           array = $.extend(true, {}, temp);
           _result.filters.samples = array;
         }
       });
+
+      if (vm_.customfilter.sampleUids.length > 0
+        || vm_.customfilter.patientUids.length > 0) {
+        var type = vm_.customfilter.type === 'sample' ? 'samples' : 'patients';
+        var uidsType = type === 'samples' ? 'sampleUids' : 'patientUids';
+        _result.filters[type][vm_.customfilter.id] =
```
kalletlak, line 807: since `vm_.customfilter.id` going to be the same for both sample and patient we could make it as `selectById`

app/scripts/session/sessionUtil.js
zhx828: JJ and I think email address would be useful when sharing the VS

app/scripts/main.js
zhx828: Yeah, but want to centralize the usage. If in the future, we wanna change the id, don't need to come back modify here.

kalletlak (approved):

PR #727, opened 2018-05-02
Show notification for unknown samples in VS

kalletlak:


app/scripts/vueCore.js
```diff
@@ -47,7 +47,22 @@
               status: false,
               message: 'Failed to open the study.' + (iViz.opts.emailContact ? (' Please contact ' + iViz.opts.emailContact + '.') : '')
             },
+            unknownSamples:[]
           }, watch: {
+            unknownSamples:function(newVal){
```
zhx828: Why do we create a watch for this? This will only be initialized once.

zhx828 (changes requested):
Need to remove the watcher.

zhx828 (approved):
LGTM

PR #735, opened 2018-05-15
When excluding NA cases, exclude samples without data as well

zhx828:
This is related to https://github.com/cBioPortal/cbioportal/issues/4094


kalletlak (approved):

jjgao:
@kalletlak could you try to make this available in the portal repo soon?

kalletlak:
@jjgao there is another issue. survival curve shows cases even with excluding case with NA data. This happens when the `Samples per Patient > 1`. In this case once sample has NA data and other is not NA. But overall considers that patient has non NA data. Will consider this as another issue and merge this PR.

